PR #23, opened 2018-08-31
Update worker config

pragnagopa:


ahmelsayed:
Can you change worker.py back to start.sh? We use that to pick the right venv on startup then just call worker.py

PR #29, opened 2018-09-27
Add ability to build locally

ankitkumarr:


host/2.0/stretch/amd64/python.Dockerfile
```diff
@@ -1,7 +1,8 @@
+ARG BASE_IMAGE=mcr.microsoft.com/azure-functions/base
 ARG BASE_IMAGE_TAG=dev
```
ahmelsayed: lets also change `BASE_IMAGE_TAG` to `2.0` by default

ankitkumarr: Sounds good. I will update that.

PR #31, opened 2018-10-05
Fixes #28 - Add python envs

### LINKED ISSUE #28: Minor update to Python image
As per [official Python docker image](https://github.com/docker-library/python/blob/master/3.6/stretch/slim/Dockerfile), could we add some important environment variables:
```
ENV LANG C.UTF-8
ENV PYTHON_VERSION 3.6.6
ENV PYTHON_PIP_VERSION 18.0
```

Also, while at it, why don't we update to version 3.6.6. I have been doing some tests and, so far, not a single problem.

ankitkumarr:


ahmelsayed:
Looks good to me, but did you check with @asavaritayal regarding updating python version from 3.6.4 to 3.6.6?

PR #34, opened 2018-10-10
Modify start.sh to use the bundled python worker

ankitkumarr:
This change is part of the PR - https://github.com/Azure/azure-functions-core-tools/pull/792

I was wondering if there's a specific reason we use a venv in the container? I have removed it in this PR, but I can add it back if we would still need it.

### DIFF
--- a/host/2.0/stretch/amd64/python-context/start.sh
+++ b/host/2.0/stretch/amd64/python-context/start.sh
@@ -3,8 +3,52 @@
 # Directory name for start.sh
 DIR="$(dirname $0)"
 
+# If we're not in a virtual environment, check if either:
+#   $AZURE_FUNCTIONS_VIRTUAL_ENVIRONMENT is set, use it for venv
+#   if it's not set and there is a default venv, activate it
+#   else use the container python
+if [ -z "$VIRTUAL_ENV"]
+then
+    # Determining the virtual environment entry point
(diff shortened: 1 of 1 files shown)


host/2.0/stretch/amd64/python-context/start.sh
```diff
@@ -3,47 +3,8 @@
...
-    echo "SKIP_PYTHONPATH_UPDATE == $SKIP_PYTHONPATH_UPDATE"
-fi
-
 echo "python == $(which python)"
 echo "PYTHONPATH == $PYTHONPATH"
 
 echo "starting the python worker"
-python $DIR/worker.py $@
\ No newline at end of file
+$HOME/site/wwwroot/worker/worker $@
```
ahmelsayed: This will break existing applications that don't yet have the bundled worker. I'd rather add an if condition at the top for this file 

```
if [ -f $HOME/site/wwwroot/worker/worker_bundle ];
then 
....
```

also maybe rename it to `worker_bundle` so it;s a bit more unique. I'd also run a `chmod +x` on it to make sure the x permissions weren't dropped for whatever reason.

ahmelsayed: basically if an app was published with an older cli, it should continue to work after this change.

PR #44, opened 2018-10-19
Adding nodejs image for nanoserver-1803

divyagandhii:


host/2.0/nanoserver-1803/node.Dockerfile
```diff
@@ -0,0 +1,22 @@
+# escape=`
+ARG BASE_IMAGE=mcr.microsoft.com/powershell
+ARG BASE_IMAGE_TAG=6.1.0-nanoserver-1803
+
+#Run-time image
+ARG RUN_TIME_IMAGE=mcr.microsoft.com/azure-functions/dotnet
+ARG RUN_TIME_IMAGE_TAG=2.0-nanoserver-1803
+
+FROM ${BASE_IMAGE}:${BASE_IMAGE_TAG} AS installer-env
```
ahmelsayed: for the installer image I think it'll be simpler to just specify the actual values directly without using a build arg. The reason we use build args in other images is so when we're building all the images on the CI server, we can override the default base image value with the one that was just built. That way the CI server could build a new base image, then all the ones that depend on it and push them as one command. But since that doesn't impact the `installer-env`, we can just specify it in the `FROM` line directly.

Issue #43, opened 2018-10-17
install msodbcsql in the python image


ahmelsayed:

ahmelsayed:
Fixed in #45

thanks @andrebriggs !

PR #45, opened 2018-10-23
Adding MS SQL ODBC drivers to Python image

andrebriggs:
Adding support for MS SQL 17 ODBC drivers. Address the issue described [here](https://github.com/Azure/azure-functions-docker/issues/43)

### DIFF
--- a/host/2.0/stretch/amd64/python.Dockerfile
+++ b/host/2.0/stretch/amd64/python.Dockerfile
@@ -28,7 +28,7 @@ RUN apt-get update && \
 # Support for MSSQL ODBC
 RUN apt-get update && apt-get install -my wget gnupg
 RUN curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
-RUN curl https://packages.microsoft.com/config/ubuntu/16.04/prod.list > /etc/apt/sources.list.d/mssql-release.list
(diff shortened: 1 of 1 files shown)


host/2.0/stretch/amd64/python.Dockerfile
```diff
@@ -20,8 +21,29 @@ RUN wget https://github.com/Azure/azure-functions-python-worker/archive/$WORKER_
...
 
+RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
 RUN apt-get update && \
-    apt-get install -y git make build-essential libssl-dev zlib1g-dev libbz2-dev \
+    apt-get install apt-transport-https
+
+# Support for MSSQL ODBC
+RUN apt-get update && apt-get install -my wget gnupg
+RUN curl https://packages.microsoft.com/keys/microsoft.asc | apt-key add -
+RUN curl https://packages.microsoft.com/config/ubuntu/16.04/prod.list > /etc/apt/sources.list.d/mssql-release.list
```
ahmelsayed: the image is actually based on Debian 9 (stretch), not Ubuntu 16.04. I know they are similar, but we should probably be using `https://packages.microsoft.com/config/debian/9/prod.list` instead.

andrebriggs: Cool, I'll test that and commit

PR #50, opened 2018-11-02
Add a docker image to allow caching of python packages

ankitkumarr:
Do not merge- This needs to be discussed and tested first

I did not update the mesh image yet, because we can perform tests directly in the new image first.

Reference - https://github.com/Azure/azure-functions-core-tools/pull/843


ankitkumarr:
@ahmelsayed, closing this and the associated core-tools PR as we will likely not be going forward this route.

PR #62, opened 2019-01-10
Update to Python 3.7.0

maiqbal11:
Updating Docker to use Python 3.7.0. @elprans, @asavaritayal, this is not the same as the latest "maintenance" release 3.7.1. Defaulting to 3.7.0 as that seems like the more stable version.


elprans:
> Defaulting to 3.7.0 as that seems like the more stable version.

@maiqbal11  3,7.2 is the latest version, please use that.  3.7.0 is, by definition, *not* the most stable version of 3.7, as 3.7.x are bugfix releases.

ahmelsayed:
in #65

PR #66, opened 2019-01-24
Fix PYTHONPATH

ahmelsayed:


ankitkumarr:
Actually if you also add `/root/.pyenv/versions/3.6.6/lib/python3.6` in the beginning of the `PYTHONPATH`, it does solve the issue of python worker using system dependencies brought in by the users.

ahmelsayed:
Great! I updated the PR

PR #71, opened 2019-01-29
Add java.Dockerfile for alpine

yamachu:
ref: https://github.com/Azure/azure-functions-docker/issues/70

Hi, thanks for awesome docker images 👍 
I thought development would become more easy with the official docker image so I made it.

It might be better that adding test code, but I could not push Java test functions to https://functionstests.blob.core.windows.net/public/docker/~~ .
So I do not have a test code, sorry.


ahmelsayed:
Thanks @yamachu! Can you add it to the build script like this https://github.com/Azure/azure-functions-docker/blob/master/host/build.vsts-ci.yml#L184-L203? I'll take care of the test

PR #75, opened 2019-02-13
Add appservice images with ssh-server and debuggers

### LINKED ISSUE #74: Create ssh enabled images for use in AppService
The changes needed for .Net are 
https://github.com/Azure/azure-functions-docker/commit/c6beee8c5a38bd3f1d44390a6bb6de72864f5d34

(For node we might need to launch the node process in debug mode additionally)

We will need to have separate ssh enabled images ( dotnet , node , python to begin with) to avoid security issue with having public username/password in the dockerfile. When used in appservice this is ok since the port is only exposed through kudu which requires auth.

The images will be named with -xdebug as part of their name (Ex: appsvc/php:5.6.38-apache-xdebug_1810291954). When remote debugging is enabled on a site , LWAS will pick the -xdebug image instead of regular image.

Matching launch.json config to use with vs code

{
            "name": ".NET Core SSH Attach",
            "type": "coreclr",
            "request": "attach",
            "processId": "${command:pickRemoteProcess}",
            "pipeTransport": {
                "pipeProgram": "ssh",
                "pipeArgs": [ "-T", "root@127.0.0.1", "-p", "PORT_NUMBER", "-i", "PATH_TO_SSHPRIVATEKEY" ],
                "debuggerPath": "/root/vsdbg/vsdbg",
                "pipeCwd": "${workspaceRoot}",
                "quoteArgs": true
            }
        }
debugger path is preset based on dockerfile debugger installation location.
PORT_NUMBER is from
az webapp remote-connection create -g RESOURCE_GROUP -n APP_NAME -p PORT_NUMBER

ahmelsayed:
Closes #74

### DIFF
--- a/host/2.0/stretch/amd64/appservice/python.Dockerfile
+++ b/host/2.0/stretch/amd64/appservice/python.Dockerfile
@@ -1,0 +1,15 @@
+ARG BASE_IMAGE=mcr.microsoft.com/azure-functions/python:2.0
+FROM ${BASE_IMAGE}
+
+EXPOSE 2222 80
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends openssh-server dialog && \
+    echo "root:Docker!" | chpasswd
+
+COPY sshd_config /etc/ssh/
+COPY start.sh /azure-functions-host/
+
+RUN chmod +x /azure-functions-host/start.sh
+
+ENTRYPOINT ["/azure-functions-host/start.sh"]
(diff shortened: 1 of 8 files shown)


host/2.0/stretch/amd64/appservice/python.Dockerfile
```diff
@@ -0,0 +1,15 @@
+ARG BASE_IMAGE=mcr.microsoft.com/azure-functions/python:2.0
+FROM ${BASE_IMAGE}
+
+EXPOSE 2222 80
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends openssh-server dialog && \
+    echo "root:Docker!" | chpasswd
```
yiliaomsft: You can remove the user name and password now, with the latest az webapp ssh command, there is no need for the user/password set in the container.

ahmelsayed: does that also work for remote debugging and kudu ssh?

rramachand21: You need the password - az webapp ssh hides the username password from the customer but its still required.

PR #76, opened 2019-02-20
Update python version to 3.6.8

ahmelsayed:


host/2.0/stretch/arm32v7/python.Dockerfile
```diff
@@ -7,7 +7,7 @@ ENV WORKER_TAG=1.0.0b3 \
     AZURE_FUNCTIONS_PACKAGE_VERSION=1.0.0b3 \
     LANG=C.UTF-8 \
     PYTHON_VERSION=3.7.2 \
```
maiqbal11: Is this a remnant from some older PR when tried switching to 3.7?

maiqbal11: Should this also be 3.6.8?

ahmelsayed: This is for an arm32v7 image. It's really irrelevant since I haven't updated grpc for it. It won't work anyway.

PR #81, opened 2019-02-27
Update to 2.0.12332

ahmelsayed:


host/2.0/stretch/amd64/mesh.Dockerfile
```diff
@@ -3,8 +3,8 @@ ARG BASE_IMAGE=mcr.microsoft.com/azure-functions/python:2.0
 FROM microsoft/dotnet:2.2-sdk AS installer-env
 
 ENV PublishWithAspNetCoreTargetManifest=false \
-    HOST_VERSION=2.0.12310 \
-    HOST_COMMIT=80f843c0f4cf778191f1d896be9f9007d5abe289
+    HOST_VERSION=2.0.12333 \
```
pragnagopa: 12332

ahmelsayed: I usually have the mesh build +1 the build number because it builds from 1 commit higher than the actual release. The actual image will still be tagged with 2.0.12332, but the runtime version inside is +1 to make it a bit easier to tell

pragnagopa:
Runtime release is complete. Please update docker

PR #85, opened 2019-03-18
Reduce layers in python.Dockerfile

jhowardmsft:
Signed-off-by: John Howard <jhoward@microsoft.com>

@ahmetsayed Following on from internal email. This will remove 3 layers from the base functions image used to build `mcr.microsoft.com/azure-functions/mesh:2.0.12332`.  Removing 3 layers will remove 3 hot-adds of VHDs to the UtilityVM which should reduce container start time by circa 100ms.

@jterry75 FYI

### DIFF
--- a/host/2.0/stretch/amd64/python.Dockerfile
+++ b/host/2.0/stretch/amd64/python.Dockerfile
@@ -4,7 +4,15 @@ FROM ${BASE_IMAGE} as runtime-image
 FROM microsoft/dotnet:2.2-aspnetcore-runtime
 
 ENV LANG=C.UTF-8 \
-    ACCEPT_EULA=Y
+    ACCEPT_EULA=Y \
+    PYTHON_VERSION=3.6.8 \
+    PYTHON_PIP_VERSION=19.0 \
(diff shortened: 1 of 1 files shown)


jterry75:
LGTM (not a maintainer)

The start ENV block is still 1 full "empty" layer right? But we just don't really have a way to avoid that?

jhowardmsft:
> The start ENV block is still 1 full "empty" layer right? But we just don't really have a way to avoid that?

I'm looking into that, but yes, correct at the moment.

jhowardmsft:
@ahmetsayed Could we get a new release of the whole chain pushed to MCR (if not already - not sure what happens now this is merged) so that the perf team can validate the end-to-end perf gain?

(Would be helpful if you could reply appropriately on the current MSFT internal email thread about this).

Thanks :)

jhowardmsft:
And to clarify - the actual measured increase on my test machine was 533ms for start time, far more than my 100ms estimate.

ahmelsayed:
Thank you @jhowardmsft for your contribution :)

PR #102, opened 2019-05-04
Update to 2.0.12452

soninaren:


host/2.0/stretch/amd64/powershell.Dockerfile
```diff
@@ -20,6 +20,14 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     HOME=/home \
     FUNCTIONS_WORKER_RUNTIME=powershell
 
+RUN apt-get update && \
+    apt-get install -y gnupg wget unzip && \
+    wget https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/1.0.0/Microsoft.Azure.Functions.ExtensionBundle.1.0.0.zip && \
+    mkdir -p /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/1.0.0 && \
+    unzip /Microsoft.Azure.Functions.ExtensionBundle.1.0.0.zip -d /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/1.0.0
+
+RUN rm -f /Microsoft.Azure.Functions.ExtensionBundle.1.0.0.zip
```
ahmelsayed: move this to the same RUN command as above. Otherwise the deleting won't actually save any size in the final image.

soninaren: Should i make the same change in base image as well?

PR #107, opened 2019-05-23
Set languageWorker debug values for appservice

ahmelsayed:


host/2.0/stretch/amd64/appservice/start.sh
```diff
@@ -1,13 +1,18 @@
...
 fi
 
 if [ -z $SSH_PORT ]; then
-  SSH_PORT=2222
+  export SSH_PORT=2222
+fi
+
+if [ "$APPSVC_REMOTE_DEBUGGING" == "TRUE" ]; then
+    export $languageWorkers__node__arguments="--inspect=0.0.0.0:$APPSVC_TUNNEL_PORT"
+    export $languageWorkers__python__arguments="-m ptvsd --server --port $APPSVC_TUNNEL_PORT"
```
maiqbal11: `ptvsd` is not a default modue in Python so we'll need to install it before we can do this.

ahmelsayed: Thanks Maheer for catching that. I added it to the image.

PR #108, opened 2019-05-24
Replace apt-get with apk in alpine images

eduardomourar:
This fixes the following error while building the alpine images:
> apt-get: not found


ahmelsayed:
Thanks @eduardomourar

PR #115, opened 2019-06-21
Update to 2.0.12543

mathewc:


maiqbal11:
@mathewc @ahmelsayed, pulled down the latest image for validation. Is it intended that `mesh.Dockerfile` has `2.0.12544` instead of `2.0.12543`?

mathewc:
Yes - Ahmed says that's the way we do it (+1 for mesh)

ahmelsayed:
@maiqbal11 Yes, I usually have the version slightly different because I would like to be able to tell if a custom image has been based on the wrong image. Also sometimes there are Mesh only releases, or bug fixes, so I always keep the version at least 1 ahead.

PR #117, opened 2019-06-25
Create package folders for importing

maiqbal11:
Imports fail if a started Python process is not pointing to created directories which are later populated with modules. Adding a fix to create the relevant directories on startup.

### DIFF
--- a/host/2.0/stretch/amd64/python-context/start.sh
+++ b/host/2.0/stretch/amd64/python-context/start.sh
@@ -6,6 +6,10 @@ DIR="$(dirname $0)"
 CUSTOM_PACKAGES_PY36="$HOME/site/wwwroot/.python_packages/lib/python3.6/site-packages"
 CUSTOM_VENV_PACKAGES_PY36="$HOME/site/wwwroot/worker_venv/lib/python3.6/site-packages"
 
+# Needed to import Python packages 
+mkdir -p "$CUSTOM_PACKAGES_PY36"
+mkdir -p "$CUSTOM_VENV_PACKAGES_PY36"
+
 export PYTHONPATH=$CUSTOM_PACKAGES_PY36:$CUSTOM_VENV_PACKAGES_PY36:$PYTHONPATH
 
 echo "python == $(which python)"
--- a/host/2.0/stretch/amd64/python-context/start.sh
+++ b/host/2.0/stretch/amd64/python-context/start.sh
@@ -3,15 +3,6 @@
 # Directory name for start.sh
 DIR="$(dirname $0)"
 
-CUSTOM_PACKAGES_PY36="$HOME/site/wwwroot/.python_packages/lib/python3.6/site-packages"
-CUSTOM_VENV_PACKAGES_PY36="$HOME/site/wwwroot/worker_venv/lib/python3.6/site-packages"
-
-# Needed to import Python packages 
-mkdir -p "$CUSTOM_PACKAGES_PY36"
-mkdir -p "$CUSTOM_VENV_PACKAGES_PY36"
-
-export PYTHONPATH=$CUSTOM_PACKAGES_PY36:$CUSTOM_VENV_PACKAGES_PY36:$PYTHONPATH
-
 echo "python == $(which python)"
 echo "PYTHONPATH == $PYTHONPATH"
 


maiqbal11:
@ankitkumarr, @Hazhzeng, this should take care of the import issue. Just need to create the folder to make the Python import machinery play nice.

ankitkumarr:
@maiqbal11, I wonder if this can cause issues if the worker is restarted at some point. I am not sure if that can happen without a host restart.

Imagine this scenario-
1. Running node function app with files mounted in /home/site/wwwroot
2. All the workers are restarted (placeholder)
3. mkdir may fail because wwwroot is read-only

If this is possible, we might want to test it out. I think this case might be possible in an App Service plan.

ankitkumarr:
Additionally, if we can handle the `PYTHONPATH` update in the worker, we won't need `start.sh` anymore. We will directly be able start the python worker (without going through bash), consistent with other workers.

maiqbal11:
> Additionally, if we can handle the `PYTHONPATH` update in the worker, we won't need `start.sh` anymore. We will directly be able start the python worker (without going through bash), consistent with other workers.

@ankitkumarr, I tested the approach where we handle the `PYTHONPATH` update in the worker by adding the package paths to `sys.path`. This won't work unless the directory structure is already present. There's no way around actually creating the directories - even if they are empty at the start. At that point, it won't make a difference whether the handling happens via `sys.path` or `PYTHONPATH`.

maiqbal11:
> 2. All the workers are restarted (placeholder)

The entire point of having placeholders is that workers aren't restarted. We simply reload the environment into the existing existing process so that we don't incur the cost of having the worker start when we execute the function. 

> 3. mkdir may fail because wwwroot is read-only

This isn't a concern if the mkdir has already succeeded once. We don't need to create it every time. Only when we initially start executing. If wwwroot is readonly then we wouldn't be able to extract the host contents into it, leading to larger issues.

Hazhzeng:
There is an issue here, when we are enabling squashfs mounting, the folder under `wwwroot/` needs to be empty. Otherwise, mount will fail.

maiqbal11:
> There is an issue here, when we are enabling squashfs mounting, the folder under `wwwroot/` needs to be empty. Otherwise, mount will fail.

We are definitely going to need to the directory to be present when we are starting the Python worker in placeholder mode. It would be better if the Host just deletes the directory if it is present before doing mounting.

ankitkumarr:
> > Additionally, if we can handle the `PYTHONPATH` update in the worker, we won't need `start.sh` anymore. We will directly be able start the python worker (without going through bash), consistent with other workers.
> 
> @ankitkumarr, I tested the approach where we handle the `PYTHONPATH` update in the worker by adding the package paths to `sys.path`. This won't work unless the directory structure is already present. There's no way around actually creating the directories - even if they are empty at the start. At that point, it won't make a difference whether the handling happens via `sys.path` or `PYTHONPATH`.

@maiqbal11, I was thinking more on the lines of updating `sys.path` when the call for function indexing is received by the worker, i.e. when the directory is actually present. That way we only update the path once user packages arrive, and not have to create a fake directory.

ankitkumarr:
> The entire point of having placeholders is that workers aren't restarted. We simply reload the environment into the existing existing process so that we don't incur the cost of having the worker start when we execute the function.

Yeah, but if the worker crashes due to some reason, the host would still need to restart it, I would imagine? Then it would run `start.sh` again. Now, I don't know if a worker restart always means the host itself would restart, but we should verify that.

> 
> This isn't a concern if the mkdir has already succeeded once. We don't need to create it every time. Only when we initially start executing. If wwwroot is readonly then we wouldn't be able to extract the function contents into it, leading to larger issues.

When we mount the archive to wwwroot, we make it readonly. If worker restarts after that, it would try and do `mkdir -p`, which would be fine if the directory is already there. 

Another concern is that, in placeholder, we start all the workers even if the language is not `python`. So, for a node app, python worker would create that dummy folder in wwwroot. We probably don't want to do that as wwwroot should only contain user payload.

Sorry, just thinking out loud to make sure we don't regress other scenarios.

ankitkumarr:
@maiqbal11, I thought about this, and the simplest and the best solution would be to leave all things as is and add this line
```
sys.path_importer_cache.clear()
```
in the [Python worker here](https://github.com/Azure/azure-functions-python-worker/blob/9c3f0576f877496c6eaa0bc377eef25d971667dc/azure/functions_worker/loader.py#L65). That will clear the caching for all the import loaders and things will work properly.

maiqbal11:
Closing in favor of https://github.com/Azure/azure-functions-python-worker/pull/473

PR #121, opened 2019-07-08
Add kudulite in build and release process

Hazhzeng:
### Update kudulite in build.sh
1. We can build a kudulite image with `./build.sh kudulite <tagname>`
2. The temporary built image will always be mcr.microsoft.com/azure-functions/kudulite:dev
3. Every release will tag the built image into
  - mcr.microsoft.com/azure-functions/kudulite:latest
  - mcr.microsoft.com/azure-functions/kudulite:$(ReleaseVersion)

### Miscellaneous
Update oryx build to 20190704.2

### DIFF
--- a/kudulite/Dockerfile
+++ b/kudulite/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/oryx/build:20190704.2 as main
+FROM mcr.microsoft.com/oryx/build:20190730.1 as main
 ARG BRANCH
 ENV DEBIAN_FRONTEND noninteractive
 
--- a/kudulite/Dockerfile
+++ b/kudulite/Dockerfile
@@ -1,5 +1,5 @@
-FROM oryxprod/build:20190521.9 as main
-
+FROM mcr.microsoft.com/oryx/build:20190704.2 as main
+ARG BRANCH
 ENV DEBIAN_FRONTEND noninteractive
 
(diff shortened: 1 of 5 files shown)


kudulite/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM oryxprod/build:20190521.9 as main
+FROM mcr.microsoft.com/oryx/build:20190704.2 as main
```
Hazhzeng: We are already using this version of oryx build in HK.

build.sh
```diff
@@ -134,6 +134,28 @@ function build_all_stretch_slim {
...
+
+    if [ -z "$tag" ]; then
+        tag=latest
+    fi
+
+    current_image=$ACR/$ACR_NAMESPACE/$name:$tag
+    echo -e "${CONSOLE_BOLD}${COLOR_GREEN}: Building $current_image ${CONSOLE_RESET}"
+    docker build -t $current_image -f $base_dir/Dockerfile $base_dir
+
+    if [[ "$tag" == 1\.*\.* ]]; then
```
ahmelsayed: today that tag comes from the release on github. Do you plan to cut releases with 1.0.xxx for kudu separately?

ahmelsayed: oh sorry I think you're passing that into your bash function. never mind.

PR #155, opened 2019-10-18
Add python worker to the python images

ahmelsayed:


maiqbal11:
@ahmelsayed, we shouldn't need to copy the worker anymore since it comes packaged with the host. That's the NuGet change that I had made. I will take a look to see why it's failing to start.

ahmelsayed:
The base image [removes the workers folder from under](https://github.com/Azure/azure-functions-docker/blob/master/host/2.0/stretch/amd64/base.Dockerfile#L12) `/azure-functions-host` then each language image brings only its worker ([node](https://github.com/Azure/azure-functions-docker/blob/master/host/2.0/stretch/amd64/node10.Dockerfile#L18), [java](https://github.com/Azure/azure-functions-docker/blob/master/host/2.0/stretch/amd64/java8.Dockerfile#L12)). The workers are few hundred mbs each, so this cuts down the image size significantly.

maiqbal11:
I see. I believe this copying is applicable for Dedicated/Premium images? Looks like Mesh will have the full set of workers from its own extraction of the Host.

ahmelsayed:
Yep, we need all the workers anyway so there is no need for this trick.

maiqbal11:
Got it 👍 I also know the fix for the dependencies not being resolved. We need to add an environment variable that tells the Host what runtime version we are using for Python. I can push it to this PR and we can take both at the same time?

ahmelsayed:
sure, I'll merge this and send your PR when you're ready, then I'll update the release tag to have both changes and kick off another build. Thanks @maiqbal11!

PR #157, opened 2019-10-18
Add runtime versions for 3.6/3.7

maiqbal11:
@ahmelsayed, this should fix the resolution of dependencies. If you want to try it out with your own image in the  meantime to unblock, you can run with `-e "FUNCTIONS_WORKER_RUNTIME_VERSION=3.7"`.


ahmelsayed:
👍 :shipit:

PR #168, opened 2019-10-28
Revert "Revert python worker changes in 3914651beafd1eceec (#161)"

maiqbal11:
This reverts commit 12336c6234c3c843ac1a18ffb7234978caf605cc. Adding back Python worker run path changes for the next host release.


maiqbal11:
This would be needed to leverage the Python binaries being packaged with the Host.

PR #169, opened 2019-10-30
Add a new tag for dotnet appservice images similar to other runtimes

ankitkumarr:
Should produce a tag similar to:
`2.0-dotnet2-appservice`

### DIFF
--- a/build/dotnet.sh
+++ b/build/dotnet.sh
@@ -49,6 +49,7 @@ function purge {
 
 function tag_push {
   # push default dotnet:$MAJOR_VERSION and dotnet:$MAJOR_VERSION-appservice images
+  DOTNET_MAJOR_VERSION=${MAJOR_VERSION%%.*}
   docker pull "${REGISTRY}dotnet:${RELEASE_VERSION}"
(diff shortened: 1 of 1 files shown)


pragnagopa:
Can you please add description on why this change is needed

ankitkumarr:
Sure!

Today, all other runtimes (except dotnet) have image name/tag for appservice images as follows --
```
- mcr.microsoft.com/azure-functions/python:2.0-python3.6-appservice
- mcr.microsoft.com/azure-functions/node:2.0-node8-appservice
- mcr.microsoft.com/azure-functions/node:2.0-node8-appservice
```
which makes it easy to construct the tag from our platform on the fly with something like `:<functions-version>-<language><language-version>-appservice`.

Except for dotnet, which is 
`mcr.microsoft.com/azure-functions/dotnet:2.0-appservice`
It's still simple, we'd have to ignore the runtime version, but it makes sense to have that as part of the tag.

It would be nice to make them all consistent, so platform doesn't have to care and special case dotnet. 
I know that today functions 2.0 -> dotnet core 2, and functions 3. -> dotnet core 3. But it also makes sense to separate and specify it this way.

PR #177, opened 2019-11-26
Add the push for maven zulu image

amamounelsayed:


amamounelsayed:
@ahmelsayed The new image should be mcr.microsoft.com/azure-functions/java:2-java8-build

PR #199, opened 2020-01-06
Clean up Python context in V3 dockerfiles

Hazhzeng:
Since the v3 host now bundles with Python worker, removing Python worker checkout code in dockerfiles.

- Bump host to 3.0.13054
- Remove python worker checkout code in 3.0/buster/amd64
    - (because now the python worker is a host builtin)
- Remove python context in buster amd64


host/3.0/buster/amd64/python/python37-context/start.sh
```diff
@@ -1,15 +0,0 @@
-#! /bin/bash
-
-# Directory name for start.sh
-DIR="$(dirname $0)"
-
-CUSTOM_PACKAGES_PY37="$HOME/site/wwwroot/.python_packages/lib/python3.7/site-packages"
-CUSTOM_VENV_PACKAGES_PY37="$HOME/site/wwwroot/worker_venv/lib/python3.7/site-packages"
-
-export PYTHONPATH=$CUSTOM_PACKAGES_PY37:$CUSTOM_VENV_PACKAGES_PY37:$PYTHONPATH
```
ahmelsayed: This is no longer needed?

Hazhzeng: Yes, this is same as host/2.0/stretch.
Because we move the start code from start.sh to python worker

PR #206, opened 2020-01-16
Fix build pipeline

Hazhzeng:
1. Stop Python 3.8 build in V2.0
2. Upgrade Oryx Build image to 20200114.13 after discussing with Gabriel Castro
3. Add --no-cache flag in Kudulite build


Hazhzeng:
Move to native branch https://github.com/Azure/azure-functions-docker/pull/207

PR #207, opened 2020-01-16
Fix build pipeline

Hazhzeng:
Stop Python 3.8 build in V2.0
Upgrade Oryx Build image to 20200114.13 after discussing with Gabriel Castro
Add --no-cache flag in Kudulite build


Hazhzeng:
Yesterday's revert due to a flaky test result.

PR #223, opened 2020-03-11
Update build pipelines

ahmelsayed:
This is a fairly big refactor/reogranization of most files. This change will

- Removes `build` and `release` pipelines in devops
- Makes every `Dockerfile` self-contained. Meaning you don't need to build a base-image per scenario to build the correct docker file. The original approach reduces number of times we need to rebuild the runtime from source, but adds a complicated dependency between the files since each platform (linux, alpine, arm, nanoserver) has a different base and a different set of files. So you need to know which is the right base to build, then figure out how to pass it to the correct dependent file. This change removes `base` and adds build steps to each file. 
- Introduces:
    - `v2 dotnet build`
    - `v2 node build`
    - `v2 java build`
    - `v2 python build`
    - `v2 powershell build`
    - `v3 dotnet build`
    - `v3 node build`
    - `v3 java build`
    - `v3 python build`
    - `v3 powershell build`
    - `v2 nanoserver build`
    - `v3 nanoserver build`
    - `v3 publish` _manual pipeline_
    - `v2 publish` _manual pipeline_

Build pipelines just perform the builds. `publish` pipeline takes care of tagging all the correct tags for images. 

- Build `alpine` and `arm32v7` by default for the supported languages
- Adds `update-version.ps1` and `update-version.sh` to help update the version number in all files. To update the version you can run 

```powershell
.\update-version.ps1 -MajorVersion 2.0 -FullVersion 2.0.99999
```

```bash
./update-version.sh 2.0 2.0.99999
```

- Adds a `-slim` version of all images 

**Few notes:**

- `arm32v7` builds are dotnet only. I couldn't get other languages to work on arm32v7 builds.
- `arm32v7` v3 builds don't have tests because the v3 dotnet images can't run on qemu emulator 
- `alpine` builds are `dotnet`, `powershell`, and `java` only
- `nanoserver` images are dotnet only
- `nanoserver` images `1903` and `1909` don't have tests because devops doesn't have the correct kernel version to run those images


README.md
```diff
@@ -2,7 +2,215 @@
...
+| ------------------------- | --------------- | ------------ |
+| 2.0, 2.0-python3.6           | [Dockerfile]()  | Debian 9     |
+| 2.0-slim, 2.0-python3.6-slim | [Dockerfile]()  | Debian 9     |
+| 2.0-appservice, 2.0-appservice-quickstart, 2.0-python3.6-appservice, 2.0-python3.6-appservice-quickstart | [Dockerfile]() | Debian 9  |
+| 2.0-python3.6-buildenv       | [Dockerfile]()  | Debian 9     |
+| 2.0-python3.7                | [Dockerfile]()  | Debian 9     |
+| 2.0-python3.7-slim           | [Dockerfile]()  | Debian 9     |
+| 2.0-python3.7-appservice, 2.0-python3.7-appservice-quickstart | [Dockerfile]() | Debian 9  |
+| 2.0-python3.7-buildenv       | [Dockerfile]()  | Debian 9     |
+| 2.0-python3.8                | [Dockerfile]()  | Debian 9     |
```
Hazhzeng: we don't have python 3.8 support in v2

ahmelsayed: good catch. I copied the section from the v3 above and updated the versions.

amamounelsayed:
@ahmelsayed We do not support java on linux for v2, should we continue deploy it or just cut it from the new releases?

README.md
```diff
@@ -2,7 +2,211 @@
...
+
+#### Java
+
+`mcr.microsoft.com/azure-functions/java`
+
+Linux amd64 Tags
+
+| Tags                     | Dockerfile      | OS Version   |
+| ------------------------ | --------------- | ------------ |
+| `3.0`, `3.0-java8`           | [Dockerfile]()  | Debian 10    |
```
amamounelsayed: Really small thing there is some of them has `` and other are not.
Please can you guide how is the [Dockerfile]() works? As the link did not work for me when I tried.

ahmelsayed: Will add `` to the rest of the image names. 

And yes the link doesn't work now, I still need to put the full path from github there for it to work

ahmelsayed:
> One more thing, we are going to support java11. So I followed the same thing you create it with node and python for versioning amamounelsayed/azure-functions-docker:java11/host/3.0/buster/amd64/java@feature 

> The question is, as java11 is not ready to merge yet, should we now change the folder structure or do it when java 11 is ready?

Up to you how you wanna manage that. I split languages with more than one version into sub folders because it seemed cleaner to my eye, but I'm hoping each language owner would just manage them how they like as long as each file is self contained and can be built with 1 docker command.

amamounelsayed:
@ahmelsayed Sounds good, I do not think there will be a lot of changes so it can wait after this merge so not to hold it.

PR #230, opened 2020-04-08
No such tag 3.0-bionic-arm32v7

DenisBalan:
Needed ubuntu image, but specified here in README is not found.
Those images are prefixed as *test* and may be not stable.
whole list is here
```
https://mcr.microsoft.com/v2/azure-functions/dotnet/tags/list
```


ahmelsayed:
`3.0-arm32v7` and `3.0-bionic-arm32v7` should be there now. The builds on MCR come out of the releases, not `master`, and usually the period between the updates is short enough, but the last couple of months things got delayed.

PR #253, opened 2020-05-13
PowerShell Core Tools Dockerfile, Pipeline, and Documentation

TsuyoshiUshio:
I wrote a Core Tools Image implementation and pipeline.

* dotnet SDK 2.1 with PowerShell 6 (We plan support from PowerShell 7, however, currently Azure Functions Support 6) 
* Add test (Test AzModule is available)
* Pipeline (Build and Publish) (Build pipeline was tested on Azure DevOps)
* Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ Linux amd64 Tags
 | `3.0`, `3.0-powershell6`                       | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6.Dockerfile)            | Debian 10  |
 | `3.0-slim`, `3.0-powershell6-slim`             | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6-slim.Dockerfile)       | Debian 10  |
 | `3.0-appservice`, `3.0-powershell6-appservice` | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6-appservice.Dockerfile) | Debian 10  |
-
+| `3.0-powershell6-core-tools` | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile) | Debian 10  |
 #### Java
 
 `mcr.microsoft.com/azure-functions/java`
--- a/host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
+++ b/host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
@@ -1,0 +1,100 @@
+
+FROM mcr.microsoft.com/dotnet/core/sdk:2.1
+
+# Avoid warnings by switching to noninteractive
+ENV DEBIAN_FRONTEND=noninteractive
+
+# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
+# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
+# will be updated to match your local UID/GID (when using the dockerFile property).
+# See https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+ARG PS_VERSION=6.2.3
+ARG PS_PACKAGE=powershell_${PS_VERSION}-1.debian.9_amd64.deb
+ARG PS_PACKAGE_URL=https://github.com/PowerShell/PowerShell/releases/download/v${PS_VERSION}/${PS_PACKAGE}
+
+# Download the Linux package and save it
+ADD ${PS_PACKAGE_URL} /tmp/powershell.deb
+
+# Define ENVs for Localization/Globalization
+ENV DOTNET_SYSTEM_GLOBALIZATION_INVARIANT=false \
+    LC_ALL=en_US.UTF-8 \
+    LANG=en_US.UTF-8 \
+    # set a fixed location for the Module analysis cache
+    PSModuleAnalysisCachePath=/var/cache/microsoft/powershell/PSModuleAnalysisCache/ModuleAnalysisCache
+
+ENV FUNCTIONS_WORKER_RUNTIME=powershell
+
+# Configure apt and install packages
+RUN apt-get update \
+    && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends apt-utils dialog 2>&1 \
+    && apt install -y /tmp/powershell.deb \
+    #
+    # Verify git and needed tools are installed
+    && apt-get -y install \
+    git \
+    openssh-client \
+    less \
+    iproute2 \
+    procps \
+    curl \
+    apt-transport-https \
+    gnupg2 \
+    lsb-release \
+    # less is required for help in powershell
+    less \
+    # requied to setup the locale
+    locales \
+    # required for SSL
+    ca-certificates \
+    gss-ntlmssp \
+    && apt-get dist-upgrade -y \
+    # enable en_US.UTF-8 locale
+    && sed -i 's/# en_US.UTF-8 UTF-8/en_US.UTF-8 UTF-8/g' /etc/locale.gen \
+    # generate locale
+    && locale-gen && update-locale \
+    # remove powershell package
+    && rm /tmp/powershell.deb \
+    #
+    # Install Azure Functions and Azure CLI
+    && echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $(lsb_release -cs) main" > /etc/apt/sources.list.d/azure-cli.list \
+    && echo "deb [arch=amd64] https://packages.microsoft.com/repos/microsoft-debian-$(lsb_release -cs)-prod $(lsb_release -cs) main" > /etc/apt/sources.list.d/dotnetdev.list \
+    && curl -sL https://packages.microsoft.com/keys/microsoft.asc | (OUT=$(apt-key add - 2>&1) || echo $OUT) \
+    && apt-get update \
+    && apt-get install -y azure-cli azure-functions-core-tools-3 \
+    #
+    # Create a non-root user to use if preferred - see https://aka.ms/vscode-remote/containers/non-root-user.
+    && groupadd --gid $USER_GID $USERNAME \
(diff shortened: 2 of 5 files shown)


host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+
+# This Dockerfile adds a non-root user with sudo access. Use the "remoteUser"
+# property in devcontainer.json to use it. On Linux, the container user's GID/UIDs
+# will be updated to match your local UID/GID (when using the dockerFile property).
+# See https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USERNAME=vscode
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+
+ARG PS_VERSION=6.2.3
```
AnatoliB, line 15: I think we should install the latest LTS version (7.0.0). Please note that the version of PowerShell _supported by Azure Functions_ and the version of PowerShell _installed as an external tool_ are entirely independent. The Functions Core Tools include its own self-sufficient version, and the next release will include _both_ 6 and 7. The version installed here will _not_ be used by Azure Functions in any case: it will be used by commands and scripts invoked outside of the Functions context, so I think we should just install the latest LTS.

README.md
```diff
@@ -57,7 +57,7 @@ Linux amd64 Tags
 | `3.0`, `3.0-powershell6`                       | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6.Dockerfile)            | Debian 10  |
 | `3.0-slim`, `3.0-powershell6-slim`             | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6-slim.Dockerfile)       | Debian 10  |
 | `3.0-appservice`, `3.0-powershell6-appservice` | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6-appservice.Dockerfile) | Debian 10  |
-
+| `3.0-powershell6-core-tools` | [Dockerfile](host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile) | Debian 10  |
```
AnatoliB, line 60: Not necessarily for this PR, but we will probably want to rethink the naming convention. The "6" in the PowerShell image names will not make much sense soon, as the Core Tools will include both PS6 and PS7 worker flavors, and the image building scripts just take the entire `/workers/powershell` content, which seems to be the right thing to do. But `/workers/powershell` will have `6` and `7` subfolders in Functions 3.0, and exactly the same will happen if we want to introduce a new image for "6".

I would just drop the version from the name: `3.0-powershell-core-tools`.

Alternatively, we could remove either `/workers/powershell/6` or `/workers/powershell/7` subfolders just for the sake of having two smaller images. But this would be a strange thing to do because the PowerShell version for Function apps is supposed to be determined by site config or environment variables.

@ahmelsayed: FYI. Please let us know if you have any ideas on this.

host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
AnatoliB: @eamonoreilly @anirudhgarg What do you think?

README.md
AnatoliB: Moving forward, the versioning story for PowerShell may become more interesting. For example, it is quite possible that we will add PowerShell 7.1 side-by-side with 7.0, etc., which will force us introduce more slightly different images. This sounds like one more reason to decouple image building scripts from PowerShell versioning, and leave the versioning story up to Core Tools, Functions Host, and later perhaps encapsulate it in the PowerShell worker entirely.

host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
```diff
@@ -0,0 +1,100 @@
...
+    # [Optional] Add sudo support for the non-root user
+    && apt-get install -y sudo \
+    && echo $USERNAME ALL=\(root\) NOPASSWD:ALL > /etc/sudoers.d/$USERNAME\
+    && chmod 0440 /etc/sudoers.d/$USERNAME \
+    #
+    # Clean up
+    && apt-get autoremove -y \
+    && apt-get clean -y \
+    && rm -rf /var/lib/apt/lists/* \
+    # intialize powershell module cache
```
AnatoliB, line 82: I think we should do this _after_ installing the Az module.

I'm wondering if there is a good way of forcing cache rebuild. Years ago, the best way was to just invoke a non-existing command, like `IAmSureThisCommandDoesNotExist`: this will make PowerShell look for this command in every available module, so it will rebuild the entire cache and write it to the file within a few seconds. This is a minor optimization, but if we have 10 lines of code just for rebuilding cache anyway, we could as well rebuild it completely.

host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
TsuyoshiUshio: Hi @AnatoliB  That's make sense. Thank you for the review. I'd like to switch to 7.0.0 and test it. :) I need to learn more about PowerShell and the Language worker behavior!

README.md
TsuyoshiUshio: I love the idea of the side-by-side scenario. That helps us to provide stable versions of the images rather than overwrite the major version with new revision images. It also help customer to use different version when something happened to the new image. Let's discuss with @ahmelsayed .

host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
TsuyoshiUshio: @AnatoliB 
Agree. Clean up should be the last. Very good catch. I'll do that. 
I'm not fully understand invoke the `IAmSureThisCommandDoesNotExist` strategy. So you mean 
L83 - L92 will be 

```bash
&& pwsh -Command "IAmSureThisCommandDoesNotExist" \
```
for forcing powershell cache? I tested in locally, it seems just causing error. Maybe I am wrong. Could you tell me more about the detail?  I also double check the latest way.

AnatoliB: @TsuyoshiUshio I'm actually talking about performing the `# intialize powershell module cache` part after `# install AzModule`. Clean up can stay where it is, as it should not be affected by installing Az building the cache.

When you invoke a command, PowerShell needs to determine which module contains this command. In order to do that, it locates modules available on the file system, reads and parses module files, etc., until the command is found. This can take long time if you have many large modules installed. In order to mitigate this, PowerShell stores the metadata discovered during this search in a cache, so that next time this data is more readily available. This cache is persisted in a local file. Reading and parsing this local file is much faster than scanning all the modules again.

However, when the command is found, PowerShell stops searching. So, if you have many modules installed, but PowerShell never needed to scan them, they will not be in the cache. The idea behind invoking a non-existing command is to force PowerShell to look at each and every module and build the complete cache, so the first invocation of an Az command will be much faster. This can be achieved by trying to invoke a non-existing command, you just need to ignore the error, something like:
``` PowerShell
try { IAmSureThisCommandDoesNotExist } catch { }
```

test/index.ts
```diff
@@ -169,6 +169,13 @@ const runTest = async (data: typeof map.dotnet, envStr = "") => {
       console.error("Azure CLI Not Found.")
       process.exit(1);
     }
+    // Powershell module tests Az module
+    if (imageName.indexOf("powershell") !== -1) {
+      if (parseInt(shell.exec(`docker run ${name} pwsh -Command "(Get-InstalledModule).count"`).stdout) === 0){
```
AnatoliB, line 174: You may want to be more specific here: `Get-InstalledModule -Name 'Az'`, otherwise this code is not really checking for Az.

host/3.0/buster/amd64/powershell/powershell6-core-tools.Dockerfile
TsuyoshiUshio: @AnatoliB 
That's awesome! Great explanation! I understand now. I'll do it! You are the master of PowerShell!

README.md
ahmelsayed: one of the main motivations of having an image per version is to reduce the overall size of the image. with side-by-side support, do you mean 1 function app with both PowerShell 7.0 and PowerShell 7.1 functions that will load the different runtimes? 

The size might not be as sensitive for the core-tools images, though smaller images are much faster to pull on slower connections

AnatoliB: @ahmelsayed By side-by-side, I mean the same approach that we already use for PS6 (in fact 6.2.4) and PS7 (in fact 7.0.0), we'll just add one more item to this list. At any given time, a PS app will use exactly one runtime version, so we will never have to load them _simultaneously_. However, the core tools user expects that the app can be switched to another PS version by changing the FUNCTIONS_WORKER_RUNTIME_VERSION app setting or environment variable. If we go with separate images, the user will have to know that this is not enough, and they will need to select a different image as well. On the other hand, we can set FUNCTIONS_WORKER_RUNTIME_VERSION on the image itself, so the user will just have to select a proper image, which is good.

I understand the size point, though. Currently, the PS6 worker is ~97 MB and the PS7 worker is ~56 MB (unzipped). @TsuyoshiUshio, what is your current image size (I just want to understand the relative delta)?

So, we are having a tradeoff: increased image size vs. the overhead of maintaining a separate image per PowerShell version (we'll have to add a new image when we add a new PowerShell version, and we'll have to update the image building scripts if the directory structure within the PowerShell worker ever changes). Technically, we can go either way. I slightly prefer reducing the complexity, unless the image size increase is prohibitive.

test/index.ts
TsuyoshiUshio: Thanks. Will do!

README.md
TsuyoshiUshio: @AnatoliB 
For the current one, 3.24 GB.   For core-tools, it is mainly used for devcontainer and ci. It is slight different from other images, size doesn't matter much. However, user experience is important. In terms of the dev container, that difference might not be very big deal.  (In the near, future, people might ask us a slim image for CI though).

TsuyoshiUshio:
Move to https://github.com/Azure/azure-functions-docker/pull/272

PR #267, opened 2020-05-22
Implement core-tools nightly build pipeline

TsuyoshiUshio:
This pipeline is Nightly Build Pipeline for publishing Azure Functions Core Tools.  Include documentation for this pipeline. 

image

The pipeline has multi stage. 

## 1.  prepare
Fetch the latest version of Azure Functions Core Tools. The version will be
`TargetVersion`. (e.g. 3.0.2534)  Also this stage fetch the weekday (0-6). 
## 2. build 
Build and Test images then publish images with `PrivateVersion`.  The `PrivateVersion` will be `TargetVersion` + weekday(0-6). (e.g. 3.0.2534.3)
## 3. publish 
Pull images from Private Docker Registry with `PrivateVersion` then publish with `TargetVersion` and `MajorVersion` (e.g. 3.0) 

## Mode
Pipeline has several mode for testing Dockerfile, Pipeline, and Rollback the images in case of the images are broken. For more details, please refer the documentation. 

## How To Use and Trouble Shooting 
The document covers both scenario. 

## TODO 
PowerShell is not included.

### DIFF
--- a/core-tools-pipeline.md
+++ b/core-tools-pipeline.md
@@ -102,4 +102,9 @@ In this case, you can ignore it and wait until the apt-get azure functions core 
 ### Pipeline has got stuck
 Sometimes, I see the issue that Azure Pipelinse stop working during pulling some images. Currently, I don't know the root cause, however, as a workaround, cancel the pipeline and execute it again after waiting for a while. Then  it will success. 
 
+### Pipeline Fails without change
+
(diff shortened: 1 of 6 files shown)


ahmelsayed:
we could use one like that for publishing the final images instead of setting variables to a manual pipeline, right? It would be nice if we can automate that further to be less error prone

PR #268, opened 2020-05-22
Remove FUNCTIONS_WORKER_RUNTIME_VERSION

amamounelsayed:


Hazhzeng:
Could you give me more content on this one?

amamounelsayed:
@Hazhzeng Java 8 and java 11 does not use FUNCTIONS_WORKER_RUNTIME_VERSION, I added in case we may need it in future. But to avoid any confusion why we have it, I would like to remove it and add it back when need it.

PR #308, opened 2020-08-05
Use runtime-deps:2.1 for V2 images to get latest os updates 

balag0:
updated runtime-deps:2.2 image is no longer available.


divyagandhii:
cc @anirudhgarg .  I thought we had agreed not to move to 2.1 until we have a meeting to discuss this. This change also conflicts with my PR for dotnet runtime. I have reverted it in master.

balag0:
Thanks for reviewing @ahmelsayed. Started an email thread to make sure everyone is on the same page.

It is okay to revert this commit for now but my understanding is it shouldn't impact anything. Lets discuss on the email thread. Thanks

PR #311, opened 2020-08-06
Use runtime-deps:2.1-bionic-arm32v7 as the base image for 2.2 bionic arm32 image

balag0:
Patched mcr.microsoft.com/dotnet/core/runtime-deps:2.2-bionic-arm32v7 is no longer available.

Using the latest 2.2 version from https://dotnet.microsoft.com/download/dotnet-core/2.2
SHA512 checksum from https://dotnetcli.blob.core.windows.net/dotnet/checksums/2.2.8-sha.txt


balag0:
after chatting with @ahmelsayed we decided to update v2 arm images only since this is not used by anyone currently.
So closing this PR.

PR #320, opened 2020-08-20
Reduce KuduLite image size

Hazhzeng:
Previously our KuduLite image reaches 8.78GB and Atlas team has hard time caching our images.

### Fixes
In this PR, I made a few improvements to reduce the image size:
1. Turning on ENABLE_DYNAMIC_INSTALL, this is a Oryx feature to lazy load deployment tools (e.g. pip, npm, dotnet) only when customer triggers a build.
2. Clean up built-in SDKs by removing python, php, hugo and stripping nodejs!=9 and dotnet~2.
3. Squash the final image into a single layer.

### Miscellaneous
1. Update KuduLite website to use .NetCore 3.1

### Validation
Building the image from https://github.com/Hazhzeng/KuduLite/tree/hazeng-stage.
Uploaded the image to docker hub rogerxman/test:kudu-one and tested on private stamp.

The newly generated image now takes 3.01 GB.

### DIFF
--- a/kudulite/startup.sh
+++ b/kudulite/startup.sh
@@ -50,4 +50,5 @@ export APPDATA=/opt/Kudu/local
 cd /opt/Kudu
 
 echo $(date) running .net core
-ASPNETCORE_URLS=http://0.0.0.0:"$PORT" runuser -p -u "$USER_NAME" -- benv dotnet=3.1 dotnet Kudu.Services.Web.dll
+# TODO: This will be updated to dotnet 3.1 soon
+ASPNETCORE_URLS=http://0.0.0.0:"$PORT" runuser -p -u "$USER_NAME" -- benv dotnet=2.2 dotnet Kudu.Services.Web.dll
--- a/kudulite/startup.sh
+++ b/kudulite/startup.sh
@@ -50,4 +50,4 @@ export APPDATA=/opt/Kudu/local
 cd /opt/Kudu
 
 echo $(date) running .net core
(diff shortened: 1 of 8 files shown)


kudulite/startup.sh
```diff
@@ -50,4 +50,4 @@ export APPDATA=/opt/Kudu/local
 cd /opt/Kudu
 
 echo $(date) running .net core
-ASPNETCORE_URLS=http://0.0.0.0:"$PORT" runuser -p -u "$USER_NAME" -- benv dotnet=2.2 dotnet Kudu.Services.Web.dll
+ASPNETCORE_URLS=http://0.0.0.0:"$PORT" runuser -p -u "$USER_NAME" -- benv dotnet=3.1 dotnet Kudu.Services.Web.dll
```
sanchitmehta, line 53: Kudu for carrying out git deploys spawns a Kudu.Console process. I'm not sure if we support Git Deploys with Consumption functions but there is a change in `3.1 Migration` where for linux since Kudu is a self contained app it  would spawn `/opt/Kudu/Kudu.Console/kudu` previously this was `benv dotnet=2.2 /opt/Kudu/Kudu.Console/kudu.dll` since here it is not a self contained app that might fail

Hazhzeng: Discussed offline, Linux Consumption does not support [Git Deploy](https://docs.microsoft.com/en-us/azure/azure-functions/functions-deployment-technologies#deployment-technology-availability) and Console for now.

Hazhzeng:
CI fails due to github.com/Azure-App-Service/KuduLite master branch is still using dotnet~2

Issue #326, opened 2020-08-31
PowerShell 7 image


AdamCoulterOz:
All guidance is being given to use PowerShell 7, but currently you don't have an image available for PowerShell 7. Can you please advise?

AdamCoulterOz:
Hey @anthonychu can we get a hand with this?

anthonychu:
@stefanushinardi

AdamCoulterOz:
I have published a PR #328 which adds a powershell 7 dockerfile for Debian buster only.

TsuyoshiUshio:
I had a look the PR and @AnatoliB already mentioned AzDo pipeline files. I don't have any other comments.  
I did quick research for core tools, It already publish 7.0.3 images also DevContainer repo is already update to enable it includes pipeline.
https://github.com/microsoft/vscode-dev-containers/blob/master/containers/azure-functions-pwsh/.devcontainer/Dockerfile

AnatoliB:
The code is merged, the built images will be published with the next Functions release.

Issue #325, opened 2020-08-31
PowerShell 3.0 Buster Base Image


AdamCoulterOz:
The `mcr.microsoft.com/azure-functions/powershell:3.0` image with the Dockerfile here (azure-functions-docker/host/3.0/**buster**/amd64/powershell/powershell6.Dockerfile) is built from a base of `mcr.microsoft.com/dotnet/core/aspnet:2.2` which is actually `debian-stretch` not `debian-buster` as indicated. This needs to be updated to have a base of `mcr.microsoft.com/dotnet/core/aspnet:3.1-buster-slim`. 

Until this change is made the guidance on what container to use for PowerShell functions is broken. 

- https://hub.docker.com/_/microsoft-azure-functions-powershell
- https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-function-linux-custom-image?tabs=bash%2Cportal&pivots=programming-language-powershell

https://github.com/Azure/azure-functions-docker/blob/5b4da7a6be22d469dc8537a270c5441310d204ac/host/3.0/buster/amd64/powershell/powershell6.Dockerfile#L32-L59

AdamCoulterOz:
also this one if possible? @anthonychu

anthonychu:
@TsuyoshiUshio @AnatoliB Can you please take a look?

AdamCoulterOz:
PR #328 addresses this problem for me, but technically all the images which claim to be `buster` but are pointing at aspnet:2.1 are actually still `stretch`. To support powershell 6 you'll need a base image which is based on `buster-slim` and has both dotnet core 2 and 3 included, as powershell 6 uses dotnet core 2.

AnatoliB:
As PowerShell 6 has reached end-of-life, we want to minimize further investments in maintaining PowerShell 6 images, so I propose closing this issue. The new PowerShell 7 images are already based on `buster`, so we will encourage you to use those. Please let us know if you have any special reason to continue using PowerShell 6 images **and** have them updated to `buster`.

AdamCoulterOz:
Fine for that logic... then update the docs to not indicate there is buster support for PS6 when there isn't. But thanks for the PS7 buster update.

PR #328, opened 2020-09-01
PowerShell 7 Debian Buster Dockerfile

AdamCoulterOz:
- Followed pattern from powershell6.Dockerfile
- Updated published container to use correct base for Debian buster
- Removed PowerShell 6 support as it would fail due to lack of dotnet core 2
- Only included version 2 of the extension bundles, and copy them from the builder container rather than repeat the install on the published container

this addresses both #325  and  #326 ... not sure how it will get published to mcr / dockerhub

### DIFF
--- a/host/3.0/buster/amd64/powershell/powershell7.Dockerfile
+++ b/host/3.0/buster/amd64/powershell/powershell7.Dockerfile
@@ -1,0 +1,54 @@
+# Build the runtime from source
+ARG HOST_VERSION=3.0.14358
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS runtime-image
+ARG HOST_VERSION
+
+ENV PublishWithAspNetCoreTargetManifest false
+
+# Build WebJobs.Script.WebHost from source
+RUN BUILD_NUMBER=$(echo ${HOST_VERSION} | cut -d'.' -f 3) && \
+    git clone --branch v${HOST_VERSION} https://github.com/Azure/azure-functions-host /src/azure-functions-host && \
+    cd /src/azure-functions-host && \
+    HOST_COMMIT=$(git rev-list -1 HEAD) && \
+    dotnet publish -v q /p:BuildNumber=$BUILD_NUMBER /p:CommitHash=$HOST_COMMIT src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj -c Release --output /azure-functions-host --runtime linux-x64 && \
+    mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
+    rm -rf /root/.local /root/.nuget /src
+
+# Install additional tools needed
+RUN apt-get update && \
+    apt-get install -y gnupg wget unzip jq
+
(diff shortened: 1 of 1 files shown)


host/3.0/buster/amd64/powershell/powershell7.Dockerfile
```diff
@@ -0,0 +1,54 @@
...
+    HOST_COMMIT=$(git rev-list -1 HEAD) && \
+    dotnet publish -v q /p:BuildNumber=$BUILD_NUMBER /p:CommitHash=$HOST_COMMIT src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj -c Release --output /azure-functions-host --runtime linux-x64 && \
+    mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
+    rm -rf /root/.local /root/.nuget /src
+
+# Install additional tools needed
+RUN apt-get update && \
+    apt-get install -y gnupg wget unzip jq
+
+# Install just bundle version 2
```
AnatoliB, line 21: I don't think we should get rid of Bundles v1 at this point. Bundles v2 can contain breaking change (this is why we keep releasing v1.* and v2.* separately), so we would unnecessarily break the users trying to migrate from PS6 to PS7, and they will be forced to perform one more migration at the same time. I think we should keep v1 for now, this will also be consistent with the images for other languages.

host/3.0/buster/amd64/powershell/powershell7.Dockerfile
```diff
@@ -0,0 +1,54 @@
+# Build the runtime from source
+ARG HOST_VERSION=3.0.14358
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS runtime-image
+ARG HOST_VERSION
+
+ENV PublishWithAspNetCoreTargetManifest false
```
AnatoliB, line 6: Nitpick: the original script (powershell6.Docker) uses the `=` syntax, and the syntax without `=` is discouraged now (AFAIK). Should we fix this?

host/3.0/buster/amd64/powershell/powershell7.Dockerfile
```diff
@@ -0,0 +1,54 @@
+# Build the runtime from source
+ARG HOST_VERSION=3.0.14358
```
AnatoliB, line 2: Can be bumped to 3.0.14492 now.

AnatoliB (changes requested):
Thank you @AdamCoulterOz!
We will also need to update `host/3.0/powershell-build.yml`, `host/3.0/publish.yml`, and `README.md`.
Please let us know if you want to address this and other comments soon, or I can take it from here.

AnatoliB:
I addressed the comments and reverted some changes that do not seem to be necessary, and published another PR based on this one: https://github.com/Azure/azure-functions-docker/pull/328

AnatoliB:
The issue is addressed by https://github.com/Azure/azure-functions-docker/pull/340, so I'm closing this one.

PR #427, opened 2021-04-14
resolve NuGet issue

TsuyoshiUshio:
Fix the first issue.


balag0:
Thanks. Looks like this has been fixed in the host repo now but we dont have a v2 release since then. so cutting a new release and updating the docker files with that new tag should resolve this. But the current PR can be a temporary workaround to unblock CI if needed urgently.

https://github.com/Azure/azure-functions-host/commit/b18f11951bb095a425664f1b6121d9846c68cdba#diff-9c5952957709545aad811b400e7e516eebade1e44fc4fbc23203403ffeb92c34

PR #409, opened 2021-03-23
Updating bundle V1 to 1.8.1 and V2 to 2.6.1

alrod:
### PR information

- [x] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/development-docs/cleaning-up-commits.md).
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made and CI is passing.

PR #430, opened 2021-04-19
Extract Base Image - Node

CooperLink:
Releasing new Host versions.

---

This PR is the first in a series of PRs that refactors existing images to more consistently rely on common base layers.  
By extracting common logic to a single base image located at amd64/functions-base/functions-host.Dockerfile it should be easier for developers to make change to images without invalidating worker's docker cache. 

### PR information

### DIFF
--- a/host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
+++ b/host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
@@ -14,10 +14,10 @@ RUN BUILD_NUMBER=$(echo ${HOST_VERSION} | cut -d'.' -f 3) && \
     rm -rf /azure-functions-host/workers/powershell/7/runtimes/win* && \
     mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
     rm -rf /root/.local /root/.nuget /src
-    
+
 ARG EXTENSION_BUNDLE_VERSION
 RUN EXTENSION_BUNDLE_VERSION=${EXTENSION_BUNDLE_VERSION} && \
-    EXTENSION_BUNDLE_FILENAME=Microsoft.Azure.Functions.ExtensionBundle.1.6.0_linux-x64.zip && \
+    EXTENSION_BUNDLE_FILENAME=Microsoft.Azure.Functions.ExtensionBundle.${EXTENSION_BUNDLE_VERSION}_linux-x64.zip && \
     apt-get update && \
     apt-get install -y gnupg wget unzip && \
     wget https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION/$EXTENSION_BUNDLE_FILENAME && \
--- a/host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
+++ b/host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
@@ -1,4 +1,5 @@
 ARG HOST_VERSION=3.0.15584
+ARG EXTENSION_BUNDLE_VERSION=1.6.0
 FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS runtime-image
 ARG HOST_VERSION
 
@@ -13,8 +14,9 @@ RUN BUILD_NUMBER=$(echo ${HOST_VERSION} | cut -d'.' -f 3) && \
     rm -rf /azure-functions-host/workers/powershell/7/runtimes/win* && \
     mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
     rm -rf /root/.local /root/.nuget /src
-
-RUN EXTENSION_BUNDLE_VERSION=1.6.0 && \
+    
+ARG EXTENSION_BUNDLE_VERSION
+RUN EXTENSION_BUNDLE_VERSION=${EXTENSION_BUNDLE_VERSION} && \
     EXTENSION_BUNDLE_FILENAME=Microsoft.Azure.Functions.ExtensionBundle.1.6.0_linux-x64.zip && \
     apt-get update && \
     apt-get install -y gnupg wget unzip && \
--- a/host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
+++ b/host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
@@ -1,0 +1,31 @@
+ARG HOST_VERSION=3.0.15584
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS runtime-image
+ARG HOST_VERSION
+
+ENV PublishWithAspNetCoreTargetManifest=false
+
+RUN BUILD_NUMBER=$(echo ${HOST_VERSION} | cut -d'.' -f 3) && \
+    git clone --branch v${HOST_VERSION} https://github.com/Azure/azure-functions-host.git /src/azure-functions-host && \
+    cd /src/azure-functions-host && \
+    HOST_COMMIT=$(git rev-list -1 HEAD) && \
+    dotnet publish -v q /p:BuildNumber=$BUILD_NUMBER /p:CommitHash=$HOST_COMMIT src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj -c Release --output /azure-functions-host --runtime linux-x64 && \
+    rm -rf /azure-functions-host/workers/powershell/6/ && \
+    rm -rf /azure-functions-host/workers/powershell/7/runtimes/win* && \
+    mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
+    rm -rf /root/.local /root/.nuget /src
+
+RUN EXTENSION_BUNDLE_VERSION=1.6.0 && \
+    EXTENSION_BUNDLE_FILENAME=Microsoft.Azure.Functions.ExtensionBundle.1.6.0_linux-x64.zip && \
+    apt-get update && \
+    apt-get install -y gnupg wget unzip && \
+    wget https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION/$EXTENSION_BUNDLE_FILENAME && \
+    mkdir -p /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION && \
+    unzip /$EXTENSION_BUNDLE_FILENAME -d /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION && \
+    rm -f /$EXTENSION_BUNDLE_FILENAME && \
+    EXTENSION_BUNDLE_VERSION_V2=2.4.0 && \
+    EXTENSION_BUNDLE_FILENAME_V2=Microsoft.Azure.Functions.ExtensionBundle.${EXTENSION_BUNDLE_VERSION_V2}_linux-x64.zip && \
+    wget https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2/$EXTENSION_BUNDLE_FILENAME_V2 && \
+    mkdir -p /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2 && \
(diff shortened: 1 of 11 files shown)


host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+    git clone --branch v${HOST_VERSION} https://github.com/Azure/azure-functions-host.git /src/azure-functions-host && \
+    cd /src/azure-functions-host && \
+    HOST_COMMIT=$(git rev-list -1 HEAD) && \
+    dotnet publish -v q /p:BuildNumber=$BUILD_NUMBER /p:CommitHash=$HOST_COMMIT src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj -c Release --output /azure-functions-host --runtime linux-x64 && \
+    rm -rf /azure-functions-host/workers/powershell/6/ && \
+    rm -rf /azure-functions-host/workers/powershell/7/runtimes/win* && \
+    mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
+    rm -rf /root/.local /root/.nuget /src
+
+RUN EXTENSION_BUNDLE_VERSION=1.6.0 && \
```
TsuyoshiUshio: Nit: This line and next line, we can make 1.6.0 to ARG. It might help visibility.

CooperLink: Testing this now locally.  Will update in next commit

host/3.0/buster/amd64/functions-base/functions-host.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    HOST_COMMIT=$(git rev-list -1 HEAD) && \
+    dotnet publish -v q /p:BuildNumber=$BUILD_NUMBER /p:CommitHash=$HOST_COMMIT src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj -c Release --output /azure-functions-host --runtime linux-x64 && \
+    rm -rf /azure-functions-host/workers/powershell/6/ && \
+    rm -rf /azure-functions-host/workers/powershell/7/runtimes/win* && \
+    mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
+    rm -rf /root/.local /root/.nuget /src
+    
+ARG EXTENSION_BUNDLE_VERSION
+RUN EXTENSION_BUNDLE_VERSION=${EXTENSION_BUNDLE_VERSION} && \
+    EXTENSION_BUNDLE_FILENAME=Microsoft.Azure.Functions.ExtensionBundle.1.6.0_linux-x64.zip && \
```
TsuyoshiUshio: You can also apply ${EXTENSION_BUNDLE_VERSION} in here. :)

CooperLink: Thank you Tsuyoshi! I just went on autopilot and git pushed too early.  All readability updates are incorporated now

divyagandhisethi:
> 

Isn't all this part common in all node images as well? I believe the only step which is different is where we install node. 

---
Refers to: host/3.0/buster/amd64/node/node10/node10-appservice.Dockerfile:30 in 199ef81. [](commit_id = 199ef8102da528d554f344439c19ab69e8b0fa9f, deletion_comment = False)

host/3.0/buster/amd64/node/node14/node14-appservice.Dockerfile
```diff
@@ -1,33 +1,5 @@
...
-    EXTENSION_BUNDLE_VERSION_V2=2.4.0 && \
-    EXTENSION_BUNDLE_FILENAME_V2=Microsoft.Azure.Functions.ExtensionBundle.${EXTENSION_BUNDLE_VERSION_V2}_linux-x64.zip && \
-    wget https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2/$EXTENSION_BUNDLE_FILENAME_V2 && \
-    mkdir -p /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2 && \
-    unzip /$EXTENSION_BUNDLE_FILENAME_V2 -d /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2 && \
-    rm -f /$EXTENSION_BUNDLE_FILENAME_V2 &&\
-    find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
+FROM functionshost:${HOST_VERSION} AS runtime-image
 
 FROM mcr.microsoft.com/dotnet/core/runtime-deps:3.1
```
divyagandhisethi, line 4: Isn't all this common as well for all node images? The only step which is different where we install node.

host/3.0/buster/amd64/node/node14/node14-slim.Dockerfile
```diff
@@ -1,33 +1,5 @@
...
-    unzip /$EXTENSION_BUNDLE_FILENAME -d /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION && \
-    rm -f /$EXTENSION_BUNDLE_FILENAME && \
-    EXTENSION_BUNDLE_VERSION_V2=2.4.0 && \
-    EXTENSION_BUNDLE_FILENAME_V2=Microsoft.Azure.Functions.ExtensionBundle.${EXTENSION_BUNDLE_VERSION_V2}_linux-x64.zip && \
-    wget https://functionscdn.azureedge.net/public/ExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2/$EXTENSION_BUNDLE_FILENAME_V2 && \
-    mkdir -p /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2 && \
-    unzip /$EXTENSION_BUNDLE_FILENAME_V2 -d /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION_V2 && \
-    rm -f /$EXTENSION_BUNDLE_FILENAME_V2 &&\
-    find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
+FROM functionshost:${HOST_VERSION} AS runtime-image
```
divyagandhisethi, line 2: Is it possible to pass the version of node to install as a parameter to the base image so that node installation can also be moved up?

CooperLink: I chose to hold off of making changes like this because the base image in this commit is intended for reuse in all other languages. EDIT : Specifically keeping the base language agnostic helps to unify the base layer and increase caching.  But, I am open to discussion about changing that. I just felt that it would be cleaner to have a single base image that is common to all languages. 

It is definitely possible to pass in node versions/any other versioning as an ARG during build.  I am also looking into Tsuyoshi/Bala's suggestion of using a tool to template the images at build time rather than maintaining all of the dockerfiles.

host/3.0/buster/amd64/node/node10/node10-appservice.Dockerfile
```diff
@@ -1,33 +1,5 @@
-# Build the runtime from source
 ARG HOST_VERSION=3.0.15584
```
pragnagopa, line 1: we should remove hardcoded `HOST_VERSION` in all these files

host/3.0/buster/amd64/node/node14/node14-appservice.Dockerfile
pragnagopa: Yes, I agree this should be part of the base image

host/3.0/buster/amd64/node/node10/node10-appservice.Dockerfile
CooperLink: I am still unsure if the ENV var for Host_Version is required anywhere. I did not see anything in Host that requires it (Other than the Proto schema linked below), so I wanted to use this PR to ask if anyone objected to me removing it.

https://github.com/Azure/azure-functions-host/blob/852eed9ceef6ef56b431428a8eb31f1bd9c97f3b/src/WebJobs.Script.Grpc/azure-functions-language-worker-protobuf/src/proto/FunctionRpc.proto#L95

host/3.0/node-build.yml
```diff
@@ -29,6 +31,14 @@ steps:
     continueOnError: false
     env:
       pswd: $(dockerPassword)
+  
+  - bash: |
+      set -e
+      HOST_VERSION=3.0.15584
```
pragnagopa, line 37: Tagging @TsuyoshiUshio - for more details on what needs to be updated in the build pipelines

- `HOST_VERSION` should be an input parameter
- Need to add a new pipeline to build function base image and push it mcr

host/3.0/buster/amd64/node/node10/node10-appservice.Dockerfile
ahmelsayed: It's the easiest way for someone to tell the version of the host running in a container when exec`ing into it.

ahmelsayed:
> By extracting common logic to a single base image located at amd64/functions-base/functions-host.Dockerfile it should be easier for developers to make change to images without invalidating worker's docker cache.

I'm not sure if you mean this change enables that, or it's just the first in an effort to enable that? This change simplifies the individual docker files, with the downside of not being able to `docker build -f <file>` without building the host image itself (when not in CI), but otherwise should behave the same as before, right?

host/3.0/node-build.yml
ahmelsayed: One downside of having `HOST_VERSION` as an input parameter to the build task, rather than a part of a commit, is that it's more difficult to run CI automatically on it. You will have to either extract it from the PR branch name, or run it manually setting the value. This PR seems to still have it in all the docker files, as well as here. Not sure what the end goal is.

CooperLink:
@ahmelsayed   This is the first in a series of changes that should hopefully make it a little easier to alter docker files in the future.  There are a few examples in the current repo of dockerfile layers that are functionally identical but, issue commands in syntactically different ways.  This currently invalidates the docker cache. 

 Another example of the issues that I believe are caused by repeating the base layer code in all images is that dockerfiles come out of sync with each other regarding base versioning.  [See Alexey's comment here about dotnet images.](https://github.com/Azure/azure-functions-docker/pull/409)  I also experienced this with Python39-appservice.  These out of sync versions also invalidate the cache.

ahmelsayed:
Yes, that's true. Though keep in mind that the part abstracted in this PR was never part of the final image built. It's just an intermediate layer that builds the host, and is never part of the final image. The final image just copies the artifacts produced by it. When all the node files are built in succession on the CI, docker build step should cache that intermediate layer since it's the same across all images. 

We may need to update the README with a `how to build image` since to build a particular image locally of the host is no longer `docker built -t <name> -f <fileName> .`, it's 

```
docker build -t functionshost:3.0.99999 -f functions-host.Dockerfile .
docker build -t <name> --build-arg HOST_VERSION=3.0.99999 -f <fileName> .
```

CooperLink:
I will definitely update the readme in my next commit thank you for the suggestion. 

[I did some tests](https://hub.docker.com/repository/docker/cooperlinkwork/javatest) and now I believe that I understand your point regarding this being an intermediate image.  As long as the file system (known to docker by the sha265 hash) of the base image is the exact same, the `copy` commands that pull artifacts during final build will be cached. The syntax differences don't matter because the final image uses `copy` and not `run`.  Therefore, the most important solve would be consolidating versioning.  Currently, this commit does not really help devs consolidate versioning. <del> I will update this with a commit that removes Host_Version from the node files by inheriting from the parent run-time image. </del> That should better enable us to make sure that the images are all using the same host/bundles and are using the cache where expected. 

--edit--
Based on some research and discussion with Tsuyoshi, there is not a great way to consolidate versioning for these docker images.  I have two possible solutions in mind.  The first is that I can provide a `.env` file that includes versions.  I believe that this combined with the azure build script would allow all images to inherit some of their environment variables through that file. Meaning that `host_version` could be defined in a central location.   Alternatively, I may use the environment variable in the base image to `mkdir azure-functions-host/version/${host_version}` and then use that FS to determine the value of `host_version` in each language's dockerfile. Both of these solutions would consolidate the versioning, but I am not sure either is great.  Based on my discussion with Tsuyoshi the best way to reduce the number of repeated variables might be templating.

ahmelsayed:
Btw, this is a useful tool to look at the layers of an image https://github.com/wagoodman/dive

CooperLink:
Closing this PR as the work has been incorporated since.

PR #439, opened 2021-04-27
adding dockerfiles for initial v4 image

brettsam:
Introducing resources to build a custom container for Functions v4 (on .NET 6). We'll initially only be supporting a custom container, so keeping this small and we'll introduce other images later.


host/4.0/buster/amd64/dotnet/dotnet-inproc/dotnet.Dockerfile
```diff
@@ -0,0 +1,34 @@
+# Build the runtime from source
+ARG HOST_VERSION=3.0.15584
+FROM mcr.microsoft.com/dotnet/sdk:6.0 AS runtime-image
+ARG HOST_VERSION
+
+# Build requires 3.1 SDK
+COPY --from=mcr.microsoft.com/dotnet/core/sdk:3.1 /usr/share/dotnet /usr/share/dotnet
```
divyagandhisethi: Since you need a different version of dotnet for building and runtime, you can try something similar to here - https://github.com/Azure/azure-functions-docker/blob/dev/host/2.0/stretch/amd64/dotnet/dotnet-appservice.Dockerfile#L33
This way you can use the dotnet 3.1 image for and then add another layer for dotnet 6.0 for runtime.

brettsam: Well, it's a little trickier. We need both .NET 6 and .NET 3.1 SDKs installed in order to build -- this should be temporary as we fix up our build targets in the near future. And then we'll run on .NET 6 only.

This step is copying the sdk bits from the .NET 3.1 so that both SDKs are present during the build

host/4.0/buster/amd64/dotnet/dotnet-inproc/dotnet.Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Build the runtime from source
+ARG HOST_VERSION=4.0.0-preview.1.15799
```
brettsam: Note this isn't quite as clean as it could be, but works for now (this isn't an actual tag; I parse it below). Will follow up with improvements later.

host/4/bullseye/amd64/dotnet/dotnet-inproc/start.sh
```diff
@@ -0,0 +1,28 @@
...
+  export ASPNETCORE_URLS=http://*:$PORT
+fi
+
+if [ -z $SSH_PORT ]; then
+  export SSH_PORT=2222
+fi
+
+if [ "$APPSVC_REMOTE_DEBUGGING" == "TRUE" ]; then
+    export languageWorkers__node__arguments="--inspect=0.0.0.0:$APPSVC_TUNNEL_PORT"
+    export languageWorkers__python__arguments="-m ptvsd --host localhost --port $APPSVC_TUNNEL_PORT"
```
ankitkumarr, line 17: We probably don't need these two exports as this is for .NET image. But I know we have this is v3 as well. So up to you if want to leave it parity or remove them.

host/update-version.ps1
```diff
@@ -9,8 +9,8 @@ Param(
     $FullVersion
 )
 
-if ($MajorVersion -ne "2.0" -And $MajorVersion -ne "3.0") {
-    Write-Error "MajorVersion can only be 2.0 or 3.0. Got $MajorVersion" -ErrorAction Stop
+if ($MajorVersion -ne "2.0" -And $MajorVersion -ne "3.0" -AND $MajorVersion -ne "4.0") {
+    Write-Error "MajorVersion can only be 2.0, 3.0, or 4.0. Got $MajorVersion" -ErrorAction Stop
```
ankitkumarr: I think this should be ` -AND $MajorVersion -ne "4"` and also the error message, as the directory is now `4/bullseye/`

host/update-version.sh
```diff
@@ -8,8 +8,8 @@ function print_usage {
   echo -e "Example: $0 3.0 3.0.123525"
 }
 
-if [ "$1" != "2.0" ] && [ "$1" != "3.0" ]; then
-  echo -e "MajorVersion must be 2.0 or 3.0. Got '$1'\n"
+if [ "$1" != "2.0" ] && [ "$1" != "3.0" ] && [ "$1" != "4.0" ]; then
```
ankitkumarr: Same, should it be `[ "$1" != "4" ]`

ankitkumarr (approved):

host/update-version.ps1
brettsam: whoops, yep.

PR #438, opened 2021-04-24
Add V3 node8-appservice to support auto-upgrade

pragnagopa:
Resolves #437 

This PR is needed to map V2 apps to V3 images in Antares

PR #446, opened 2021-05-13
Publish node8-appservice image

pragnagopa:
Follow up for https://github.com/Azure/azure-functions-docker/pull/438

Publish V3 Node8 AppService Images for auto-upgrade

### DIFF
--- a/host/3.0/publish.yml
+++ b/host/3.0/publish.yml
@@ -147,6 +147,7 @@ steps:
 
   - bash: |
       set -e
+      docker pull $SOURCE_REGISTRY/node:$(PrivateVersion)-node8-appservice
       docker pull $SOURCE_REGISTRY/node:$(PrivateVersion)-node10
       docker pull $SOURCE_REGISTRY/node:$(PrivateVersion)-node10-slim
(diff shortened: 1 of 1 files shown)


host/3.0/publish.yml
```diff
@@ -157,11 +158,15 @@ steps:
       docker pull $SOURCE_REGISTRY/node:$(PrivateVersion)-node14-slim
       docker pull $SOURCE_REGISTRY/node:$(PrivateVersion)-node14-appservice
 
+      docker tag $SOURCE_REGISTRY/node:$(PrivateVersion)-node8-appservice $TARGET_REGISTRY/node:$(TargetVersion)-appservice
```
TsuyoshiUshio, line 161: This line should be removed. It looks overwrite current target version. current node:3.0 will be node10. 
https://github.com/Azure/azure-functions-docker

TsuyoshiUshio (changes requested):
You also might need to update push part. 
https://github.com/Azure/azure-functions-docker/blob/96e86ac9b7c86b976b0493ecd98349b26fa35140/host/3.0/publish.yml#L193

PR #449, opened 2021-05-21
Include V3-Node8-AppService image in build

pragnagopa:


pragnagopa:
@TsuyoshiUshio  - Please take a look.

PR #470, opened 2021-08-06
Introduce build pipeline to push KEDA images

JatinSanghvi:
### PR information

### DIFF
--- a/keda/publish.yml
+++ b/keda/publish.yml
@@ -1,0 +1,42 @@
+trigger: none 
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+
+# Bash task:https://docs.microsoft.com/azure/devops/pipelines/yaml-schema#bash.
+- bash: |
+    # Exit if any line in the script fails.
+    set -e
+
+    if [ -z "$VERSION_TAG" ];
+    then
+      echo "ERROR: VERSION_TAG is required."
+      exit 1
+    fi
+
+    SOURCE_KEDA_IMAGE=$SOURCE_REPOSITORY/kedacore/keda:$VERSION_TAG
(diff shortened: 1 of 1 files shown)


ahmelsayed:
The change looks good to me. We need to check with MCR team to see if https://github.com/microsoft/mcr/pull/1562 is ok as is, or if we will need to move that part under the `oss` folder/team there.

tomkerkhove:
Just out of curiosity - Why is there a need for this? Can't we just improve the official KEDA images or is it a "whitelisting" thing?

ahmelsayed:
Some internal Azure teams have a requirement to only pull from MCR. MCR SLA is under Azure, which makes any registry outage easier to track for Azure teams compared to external registries. It also simplifies deploying to non-connected/offline Azure environments where only MCR is replicated/available offline. The image is copied as is, so I don't think there is anything to "improve" per se in the official KEDA image. Just the choice of registry it's hosted on.

JatinSanghvi:
@ahmelsayed - Is there a way to make the PR merge build succeed? It is failing since the input variable values are not being passed to the build. Also, if in case we are using a test ACR for PR-validation build, I can remove the hardcoding of ACR location.

tomkerkhove:
> Some internal Azure teams have a requirement to only pull from MCR. MCR SLA is under Azure, which makes any registry outage easier to track for Azure teams compared to external registries. It also simplifies deploying to non-connected/offline Azure environments where only MCR is replicated/available offline. The image is copied as is, so I don't think there is anything to "improve" per se in the official KEDA image. Just the choice of registry it's hosted on.

Thanks for sharing!

ahmelsayed:
@JatinSanghvi not sure how to pass values to the auto PR run. You could set `$VERSION_TAG` default to `main`. I ran it with a value https://azure-functions.visualstudio.com/azure-functions-docker/_build/results?buildId=11662&view=results and it seems fine.

ahmelsayed (approved):

PR #488, opened 2021-08-27
Add 3.0 image push 

TsuyoshiUshio:
[v3 publish pipeline](https://azure-functions.visualstudio.com/azure-functions-docker/_apps/hub/ms.vss-build-web.ci-designer-hub?pipelineId=32&nonce=hSO3C5Yy0YsMFa4xyfhtDA%3D%3D&branch=dev) requires to run two times. If you want to publish `3.2.0` version, you need to run twice. One for `targetVersion=3.2.0` the other is `targetVersion=3.0`.   

This change assume for the V3 version, when we push the version `3.a.b` it will push `3.0` version as well. So that users can run this pipeline only one time.  It is a not good practice to publish 3.0 has the same version as 3.2.0, it is because this numbering system has been changed during the operation. v4 versioning won't have this issue.

After this PR has been merged, we need to update the following documentation. 
https://github.com/Azure/azure-functions-docker/wiki/Preview:-Release-Images-for-Linux-Dedicated-(GitHub-Action)
https://azfunc.visualstudio.com/Azure%20Functions/_release?_a=releases&view=mine&definitionId=20

FYI @CooperLink  @pgombar

### PR information

### DIFF
--- a/host/3.0/publish.yml
+++ b/host/3.0/publish.yml
@@ -49,6 +49,18 @@ steps:
       docker tag $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-slim $TARGET_REGISTRY/base:$(TargetVersion)-slim
       docker tag $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-appservice $TARGET_REGISTRY/base:$(TargetVersion)-appservice
 
+      docker tag $SOURCE_REGISTRY/dotnet:$(PrivateVersion) $TARGET_REGISTRY/dotnet:3.0
+      docker tag $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-slim $TARGET_REGISTRY/dotnet:3.0-slim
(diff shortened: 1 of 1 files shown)


CooperLink (approved):
This looks good to me. I had a few comments about the number of images that this would add to our tags. But, on second inspection it seems that all of these tags currently exist.  Before merging could you let me know if this will add any images that we would need to support in the future?  I am currently under the impression that this will only provide CI/CD updates to our existing tags.  I just wanna make sure we don't introduce any new image tags we would have to support in the far future :)

TsuyoshiUshio:
That is a good point. We don't want to support that is not important. 
We used to run this pipeline with target version e.g. `3.2.0` and `3.0`, so that I just automated the process. 
However, the officially support of the 3.0 should be limited. How about make the official support is documented version in here? https://github.com/Azure/azure-functions-docker That means, I can add Readme to state the following version is officially supported.

balag0:
@TsuyoshiUshio Do we need similar change for 4.0?

PR #503, opened 2021-09-17
Updating v4 images for preview 4.0.0-preview.4.16394

kshyju:
- Updating host version from `4.0.0-preview.3.16281` to `4.0.0-preview.4.16394`
- Updating .net sdk from `6.0.100-preview.7` to `6.0.100-rc.1`

These are the only changes in all files (30 files changed totally). I tested 3 of them locally(java, python and dotnet-inproc).

Used _**FIND and REPLACE**_ of IDE(vs code).

PR #506, opened 2021-09-17
Updating v4 images for preview 4.0.0-preview.4.16394 - Fixing few lines I missed in last PR

kshyju:
I missed a few places to update the .net runtime version in [previous PR](https://github.com/Azure/azure-functions-docker/pull/503). Fixing that in this PR. Thanks Brett for pointing those out.

### DIFF
--- a/host/4/bullseye/amd64/powershell/powershell7.Dockerfile
+++ b/host/4/bullseye/amd64/powershell/powershell7.Dockerfile
@@ -43,7 +43,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM mcr.microsoft.com/dotnet/aspnet:6.0.0-rc.1
+FROM mcr.microsoft.com/dotnet/aspnet:6.0.0-rc.1-bullseye-slim
 ARG HOST_VERSION
 
 # Powershell worker requires 3.1 for now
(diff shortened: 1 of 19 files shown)


host/4/bullseye/amd64/powershell/powershell7.Dockerfile
```diff
@@ -43,7 +43,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM mcr.microsoft.com/dotnet/aspnet:6.0.0-preview.7-bullseye-slim
+FROM mcr.microsoft.com/dotnet/aspnet:6.0.0-rc.1
```
brettsam: do these need to be `bullseye-slim`?

kshyju: @brettsam  I believe both the `6.0-bullseye-slim` and ` 6.0.0-rc.1` tags points to same image. I chose rc1 to be consistent with other instances where we used `rc.1` . Do you think we should use  `6.0-bullseye-slim` tag instead ?

https://hub.docker.com/_/microsoft-dotnet-aspnet

balag0: yeah. 6.0-bullseye-slim is much more descriptive :)

balag0: and i assume even if a rc2 image gets published, we dont have to update anything here

kshyju: Let me push another iteration to switch to `bullseye-slim` so it is consistent with previous iterations of v4 work.

kshyju: Switched back to `6.0.0-rc.1-bullseye-slim`.

brettsam (approved):

balag0 (approved):

PR #417, opened 2021-04-01
Add chrome headless dependencies to python images

ahmelsayed:
---

Add chrome headless dependencies to 

| Tags                                         |
|----------------------------------------------|
| `3.0`, `3.0-python3.6`                       |
| `3.0-appservice`, `3.0-python3.6-appservice` |
| `3.0-python3.7`                              |
| `3.0-python3.7-appservice`                   |
| `3.0-python3.8`                              |
| `3.0-python3.8-slim`                         |
| `3.0-python3.9`                              |
| `3.0-python3.9-slim`                         |

Closes #407

### PR information

- [x] The title of the PR is clear and informative.
- [x] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/development-docs/cleaning-up-commits.md).
- [x] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made and CI is passing.

PR #516, opened 2021-10-01
Templating  Dockerfiles - JAVA

CooperLink:
This PR is the first in a series of PRs to simplify the process of writing and changing our docker files.  The goal here is to introduce a common base image for the Azure Host runtime bits that can be shared across all of our images.  The method I achieve that through here is by running a script to combine our language specific images with the base image defined in this PR.

Re-using the same base image has two major benefits.  The most pronounced is to reduce the need for scripts like update-version.sh or github commits that require changes to all of our host base images.  With our current configuration, changes to the base image require over 50 file change just to maintain consistency between images.  This is dangerous as there is not an automated method to change all files at once (other than hostversion) and it is easy to miss one or two files while trying to alter all of them.  

Base image re-use hopefully reduces the need for PRs like this : https://github.com/Azure/azure-functions-docker/pull/515/files

By introducing this method, it should be much easier to change our images and maintain consistency.  A lack of consistency is an important barrier to overcome as inconsistent base images invalidate the docker cache leading to unnecessarily longer build times, quicker worker cache exhaustion, and possibly hard to track bugs.

This PR and the forthcoming PRs to apply this everywhere have limited expected impact on cache size but, will hopefully boost developer productivity, traceability and maintainability of this repo. There are numerous PRs in the history of this repo that were required to fix dockerfiles that were accidentally ignored during initial work.  This tool aims to change that. 

Another major benefit of this approach is breaking apart multi-stage builds.  This makes it easier to produce early stages as images which can allow us to cut down on image size by inspecting the runtime-image stage file system and ensuring we are only copying over what is needed. 

### PR information


host/3.0/buster/amd64/base/out/java/java11-appservice.Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1 AS runtime-image
+ARG HOST_VERSION
+
+ENV PublishWithAspNetCoreTargetManifest=false
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN BUILD_NUMBER=$(echo ${HOST_VERSION} | cut -d'.' -f 3) && \
+    git clone --branch v${HOST_VERSION} https://github.com/Azure/azure-functions-host.git /src/azure-functions-host && \
+    cd /src/azure-functions-host && \
```
pragnagopa: no need to build host again here?

host/3.0/buster/amd64/base/out/java/java11-appservice.Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+    HOST_COMMIT=$(git rev-list -1 HEAD) && \
+    dotnet publish -v q /p:BuildNumber=$BUILD_NUMBER /p:CommitHash=$HOST_COMMIT src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj -c Release --output /azure-functions-host --runtime linux-x64 && \
+    mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
+    rm -rf /root/.local /root/.nuget /src
+
+RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
```
pragnagopa: same here - extension bundle is part of host.dockerfile - remove?

host/3.0/buster/amd64/base/generate-composite.sh
```diff
@@ -0,0 +1,54 @@
...
+
+combine_dockerfile() {
+    printf $"  Combining $base with $1\n"
+    runtime=`echo $1 | cut -d'/' -f 2`
+    runtimeversion=`echo $1 | cut -d'/' -f 3`
+    printf "\t Image Runtime and Version: $runtimeversion \n"
+
+    printf $"\t Collecting global arguments in $1 to globalize them in the final dockerfile.\n"
+    globalize_args $1
+    printf '%s\n' "${gblarg[@]}" > ./temp.txt
```
TsuyoshiUshio: Nit: You can use >> as append mode. 
However, it is up to you. current code is fine as well.

host/3.0/buster/amd64/base/host.Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+RUN BUILD_NUMBER=$(echo ${HOST_VERSION} | cut -d'.' -f 3) && \
+    git clone --branch v${HOST_VERSION} https://github.com/Azure/azure-functions-host.git /src/azure-functions-host && \
+    cd /src/azure-functions-host && \
+    HOST_COMMIT=$(git rev-list -1 HEAD) && \
+    dotnet publish -v q /p:BuildNumber=$BUILD_NUMBER /p:CommitHash=$HOST_COMMIT src/WebJobs.Script.WebHost/WebJobs.Script.WebHost.csproj -c Release --output /azure-functions-host --runtime linux-x64 && \
+    mv /azure-functions-host/workers /workers && mkdir /azure-functions-host/workers && \
+    rm -rf /root/.local /root/.nuget /src
+
+RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
+    EXTENSION_BUNDLE_FILENAME=Microsoft.Azure.Functions.ExtensionBundle.1.8.1_linux-x64.zip && \
```
TsuyoshiUshio: This line should be replaced by the variables. I mean 1.8.1 part.

TsuyoshiUshio (commented):
Except for 1 comment, It looks great! Thank you for the contribution.  One thing I want to discuss is, if we should include the generated Dockerfiles or not. We saw the both approaches. If we don't include it, PR looks simple and don't see many changes. If we include it as is, it is good since we can see the generated image and which change is introduced to each Dockerfiles. however, people could confuse if they change it or not. However, if you automate it, it might doesn't matter. :)

host/3.0/buster/amd64/base/generate-composite.sh
CooperLink: I **think** that I want to not use append mode here as every individual lang/version will overwrite this file.

host/3.0/buster/amd64/base/host.Dockerfile
CooperLink: I saw this and agree. Will work on extracting all things possible to env vars in the near future.

host/3.0/buster/amd64/base/generate-composite.sh
ahmelsayed: You can do 

```
echo "${gblarg[@]}" | cat - $base $1 > ./out/${runtime}/${runtimeversion}-appservice.Dockerfile
```

To avoid the need for a tmp file.

host/3.0/buster/amd64/base/out/java/java11-appservice.Dockerfile
ahmelsayed: This is the out directory. Maybe it should be in the root of the repo so it's clearer? i.e:

```
out/host/3.0/amd64/java/java11-appservice.Docker
```

host/3.0/buster/amd64/base/temp.txt
```diff
@@ -0,0 +1 @@
+ARG JAVA_VERSION=8u252
```
ahmelsayed: remove

host/3.0/buster/amd64/java/java11/java11-composite.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+    HOST_VERSION=${HOST_VERSION}
+
+COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
+COPY --from=runtime-image [ "/workers/java", "/azure-functions-host/workers/java" ]
+COPY --from=jre [ "/usr/lib/jvm/zre-hl-tools-11-azure-amd64", "/usr/lib/jvm/zre-11-azure-amd64" ]
+
+## Using this method with ../ requires the docker build be run from the /out/ folder.
+## Alternative solution would be to copy the main files for sshd and start.sh to each lang folder
+COPY ../shared/sshd_config /etc/ssh/
+COPY ../shared/start.sh /azure-functions-host/
```
ahmelsayed: `..` doesn't really work in `Dockerfile` as you mentioned in the comment the docker build command needs to run from the `./out` dir. `../shared` will resolve to `/shared` in the current directory

host/3.0/buster/amd64/base/temp.txt
CooperLink: Going to .gitignore all the /out/files and can update the script to delete this file at the end.

host/3.0/buster/amd64/java/java11/java11-composite.Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
+
+ENV JAVA_HOME /usr/lib/jvm/zre-11-azure-amd64
```
ahmelsayed: nit: move with the list of `ENV` above

host/3.0/buster/amd64/java/java11/java11-composite.Dockerfile
CooperLink: Agreed.  Was just trying to think of a way to fix this.  I see two solutions. 
1: Move all output dockerfiles to the same base /generated/ dir
         Would allow us to just use the two config files in all the dirs instead of copying over lots of them
2: Copy configs to the language folders from a main copy in the /base/ folder
        Copying over this file into all the language folders would also simplify this problem. But, we would create lots of identical files.  Copying them all from the main copy in the /base/ folder would allow us to change all the configs at once similar to the base image.

host/3.0/buster/amd64/base/out/java/java11-appservice.Dockerfile
CooperLink: Thank you for the suggestion Ahmed.  I will probably use that path when I am finished implementing this across versions. My plan is to bump it up to all languages v3 -appservice.  Then see what I can do about the other images we have in v3. Then update it for v4. 
I wrote this down and will tag you when I have a PR that incorporates it. 

https://github.com/Azure/azure-functions-docker/issues/538

host/3.0/buster/amd64/base/temp.txt
CooperLink: Saw your comment with a command that could replace the need for this file. Incorporating.

pragnagopa:
As discussed offline - Please checkin the produced docker files into a test directory and add unit tests

host/3.0/buster/amd64/base/generate-composite.sh
CooperLink: Incorporated this in most recent commit. Thank you both

host/3.0/buster/amd64/base/temp.txt
CooperLink: Introduced the .gitignore and removed need for this file

host/3.0/buster/amd64/java/java11/java11-composite.Dockerfile
CooperLink: Chose to copy all configs over to the output folder during ./generate-composite.sh. Will push latest updates with these changes.

host/3.0/buster/amd64/node/node10/node10-composite.template
```diff
@@ -0,0 +1,42 @@
...
+
+# Chrome headless dependencies
+# https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#chrome-headless-doesnt-launch-on-unix
+RUN apt-get install -y xvfb gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 \
+    libdbus-1-3 libexpat1 libfontconfig1 libgbm1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 \
+    libgtk-3-0 libnspr4 libpango-1.0-0 libpangocairo-1.0-0 libstdc++6 libx11-6 libx11-xcb1 libxcb1 \
+    libxcomposite1 libxcursor1 libxdamage1 libxext6 libxfixes3 libxi6 libxrandr2 libxrender1 libxss1 \
+    libxtst6 ca-certificates fonts-liberation libappindicator1 libnss3 lsb-release xdg-utils wget
+
+ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
```
pragnagopa: except installing specific version of node, rest of the commands are common to all node images. This would be true for python as well -- all the common dependencies in python images except for the python version 

Does it make sense to have template per language as well?

host/3.0/buster/amd64/node/node10/node10-composite.template
```diff
@@ -0,0 +1,42 @@
...
+ARG HOST_VERSION
+
+RUN apt-get update && \
+    apt-get install -y curl gnupg && \
+    curl -sL https://deb.nodesource.com/setup_10.x | bash - && \
+    apt-get update && \
+    apt-get install -y nodejs
+
+# Chrome headless dependencies
+# https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#chrome-headless-doesnt-launch-on-unix
```
pragnagopa: Can installing `Chrome headless dependencies` move to host.dockerfile?

host/3.0/buster/amd64/node/node10/node10-composite.template
CooperLink: I can definitely do that. In the next PR that I update here I will add a function to do that based on flags for languages that can benefit from it.  For now I will keep it as just to release changes iteratively.  Will probably start by adding this for all v3 images. Then seeing if I can consolidate for node and python/ introduce vars that can be set to create each version.  Imo out of scope for this specific work. Focusing on base image extraction first. :)

host/3.0/buster/amd64/node/node10/node10-composite.template
CooperLink: Unsure, only because as far as I understand node and python are the only ones that have it.  I will investigate what I can do to consolidate this. 
https://github.com/Azure/azure-functions-docker/pull/417

You also made a comment in this PR to see if these could be moved to the main dockerfile.

host/3.0/buster/amd64/node/node10/node10-composite.template
pragnagopa: makes sense. Do track this separately - open issue / WI

host/3.0/buster/amd64/node/node10/node10-composite.template
pragnagopa: Thanks for linking earlier PR - makes sense to move this to common file as it does not seem to language specific - we can track this separately as well. 

We should move this to host.dockerfile as long as it does not increase image size drastically as Node and Python images need it anyways

CooperLink:
I had a good discussion with Bala about traceability of the dockerfiles. With the solution proposed here, we can still view git history of the files and understand changes.  But, changes to the .Dockerfiles that actually build the images that we use in production (java8-appservice.Dockerfile) would be harder to visualize. 
To fix this problem, I added a switch -r that generates images in the release folder on a branch.  I also updated the release drafter to generate those images.  So when we cut a release it will include a snapshot of the images that that version is using in production. This will also help out with a task I am doing soon to rebuild the images every two weeks.

pragnagopa:
Adding notes from offline conversation:
- Include generated docker files as " Build Artifacts"
- Upload Build Artifacts to github release page as part of release pipeline

host/3.0/buster/amd64/base/host.Dockerfile
CooperLink: Including this in a future work item and closing conversation
https://github.com/Azure/azure-functions-docker/issues/539

CooperLink:
I will be squashing and merging this branch after approval :) Just rebased.

pragnagopa (approved):
LGTM! @TsuyoshiUshio / @balag0  - please take a final look as well! @CooperLink  - please open follow up github issues for unresolved comments

balag0:
Looks great. Greatly simplifies updating host versions. One minor thing for later - if there is any guidance for new dockerfile authors / making more involved changes in existing dockerfiles, please document them somewhere.

host/3.0/buster/amd64/base/host.Dockerfile
```diff
@@ -23,7 +23,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     mkdir -p /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION && \
     unzip /$EXTENSION_BUNDLE_FILENAME -d /FuncExtensionBundles/Microsoft.Azure.Functions.ExtensionBundle/$EXTENSION_BUNDLE_VERSION && \
     rm -f /$EXTENSION_BUNDLE_FILENAME && \
-    EXTENSION_BUNDLE_VERSION_V2=2.8.3 && \
+    EXTENSION_BUNDLE_VERSION_V2=2.6.1 && \
```
TsuyoshiUshio: I'm curious about this downgrade. Could you share the context?

TsuyoshiUshio (approved):
Great Job! Thank you for the all of effort and clean bash code.

host/3.0/buster/amd64/base/host.Dockerfile
pragnagopa: good catch @TsuyoshiUshio  - this is a recent update . PR is needs to be udpated.

pragnagopa (dismissed):
@CooperLink - Lets fix the extension bundle version and merge the PR

PR #537, opened 2021-11-03
[V4] Switching from rc2 to nightly for dotnet runtime deps

kshyju:
The dotnet nightly images are [confirmed to be same bits as of the GA bits](https://twitter.com/runfaster2000/status/1455372722939310084). In this PR, I am updating the rc2 image with the nightly image for dotnet sdk & runtime deps.

We are still using rc2 of dotnet/sdk for [building & publishing](https://github.com/Azure/azure-functions-docker/blob/dev/host/4/bullseye/amd64/dotnet/dotnet-isolated/dotnet-isolated.Dockerfile#L12) the functions host project. Because some of the dependent packages for  `linux-x64` runtime (ex: `Microsoft.NETCore.App.Runtime.linux-x64`) are not yet available in public nuget feed yet. (They will be available only on Nov 8th).

There will be a follow up PR on or after 11/8th to update the remaining pieces to use GA version of dotnet sdk images.

### DIFF
--- a/host/4/bullseye/amd64/dotnet/dotnet-inproc/dotnet-appservice.Dockerfile
+++ b/host/4/bullseye/amd64/dotnet/dotnet-inproc/dotnet-appservice.Dockerfile
@@ -35,7 +35,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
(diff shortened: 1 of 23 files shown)


balag0:
Changes look good, but just curious, since we need another PR very soon to switch the build environment image you pointed out, why not just wait for the other change as well?

kshyju:
> Changes look good, but just curious, since we need another PR very soon to switch the build environment image you pointed out, why not just wait for the other change as well?

We want to test early with the GA bits changes and be ready for GA date.

pragnagopa (approved):

PR #550, opened 2021-11-09
Delete auto-upgrade images

pragnagopa:
Platform has been updated to use V3 images for `~2` apps in ANT 95. These images are not needed any more. This PR cleans up the repo and saves build times


balag0:
> Platform has been updated to use V3 images for `~2` apps in ANT 95. These images are not needed any more. This PR cleans up the repo and saves build times

Is anything changing for ~2.0 apps as well?

CooperLink:
We should probably just delete those images also. AFAIK we haven't built them in quite a while. We just build v3. v4 and previously the v2-upgrade

divyagandhisethi:
> We should probably just delete those images also. AFAIK we haven't built them in quite a while. We just build v3. v4 and previously the v2-upgrade

If we delete the ~2.0 images, then what happens to the apps which are currently pinned to v2?

divyagandhisethi (approved):

pragnagopa:
> If we delete the ~2.0 images, then what happens to the apps which are currently pinned to v2?

This PR is deleting auto-upgrade images only. `~2.0` apps will continue using V2 images

divyagandhisethi:
We can delete the corresponding devops pipelines as well.

CooperLink:
Great question Divya, luckily I just implemented this on Dedicated.  For Dedicated ~2 apps get automatically upgraded to ~3.  If someone were to select a specific version of v2 they would recieve our min supported version. I don't believe there is any way on dedicated to get a 2.0 image even if someone used DOCKER| linuxfx.
I am not 100% sure on consumption but, I am fairly confident that it works the same. I tested it a few weeks ago for my pinning work verification and found that none of the apps came up with a v2 app. 

Going to do some more investigation as well just to ensure I don't break anything before I remove the 2.0 images from the registry
EDIT: investigating the ~2.0 that Pragna mentioned. Just read through my investigation notes from before. ~2, ~2.0, 2.0 2  -> all upgrade to a v3 image

divyagandhisethi:
> Great question Divya, luckily I just implemented this on Dedicated. For Dedicated ~2 apps get automatically upgraded to ~3. If someone were to select a specific version of v2 they would recieve our min supported version. I don't believe there is any way on dedicated to get a 2.0 image even if someone used DOCKER| linuxfx. I am not 100% sure on consumption but, I am fairly confident that it works the same. I tested it a few weeks ago for my pinning work verification and found that none of the apps came up with a v2 app.
> 
> Going to do some more investigation as well just to ensure I don't break anything before I remove the 2.0 images from the registry EDIT: investigating the ~2.0 that Pragna mentioned. Just read through my investigation notes from before. ~2, ~2.0, 2.0 2 -> all upgrade to a v3 image

I see. What happens to function apps on dotnet 2.2, which can't be upgraded to dotnet 3.x? Will they be broken?

pragnagopa:
Clarifications regarding `~2.0 ` apps - for dedicated apps - the only way to pin and run V2 runtime is using `DOCKER|mcr.microsoft.com/azure-functions/dotnet:2.0.14786-appservice` --> app settings are not honored for Linux Dedicated Pinning to Azure Functions V2

pragnagopa:
> Is anything changing for ~2.0 apps as well?

No changes to V2 Apps. On Linux Dedicated / EP the only way to run V2 runtime is by setting  `siteConfig` property `linuxFxVersion` to `DOCKER|mcr.microsoft.com/azure-functions/dotnet:2.0.14786-appservice`

PR #561, opened 2021-11-18
Moving away from nightly image of dotnet/runtime-deps

### LINKED ISSUE #556: Change v4 images from nightly to stable
They were temporarily using the .NET nightlies to generate our first v4 GA image because it was the only way to get the image before Nov 8 and the .NET team was able to ensure it contained .NET 6 RTM bits. There’s no guarantee of this going forward and we need to change them to stable before next release.

/cc @CooperLink @kshyju

kshyju:
Fixes #556 

For making the V4 images ready for GA day, we [used ](https://github.com/Azure/azure-functions-docker/pull/537)the nightly images of `dotnet/runtime-deps`. In this PR, we are removing the nightly ones. Thanks to Cooper's refactoring. Only few files need the change.

Also switched the dotnet sdk image used for building code from rc2 version to RTM, since the dependent packages are now available in public nuget gallery.

### DIFF
--- a/host/4/bullseye/amd64/base/host.Dockerfile
+++ b/host/4/bullseye/amd64/base/host.Dockerfile
@@ -1,6 +1,6 @@
 ARG HOST_VERSION=4.0.1.16815
 
-FROM mcr.microsoft.com/dotnet/sdk:6.0.100-rc.2 AS runtime-image
+FROM mcr.microsoft.com/dotnet/sdk:6.0.100 AS runtime-image
 ARG HOST_VERSION
 
 ENV PublishWithAspNetCoreTargetManifest=false \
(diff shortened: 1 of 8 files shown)


pragnagopa:
> Thanks to Cooper's refactoring. Only few files need the change.

💯

pragnagopa (approved):
💯 @CooperLink  - less number of files changes in the PR

pragnagopa (changes requested):
Just realized - this needs to updated in the other language composite files too - see  https://github.com/Azure/azure-functions-docker/blob/dev/host/4/bullseye/amd64/java/java8/java8-composite.template#L4 - for example

Please do string search across the repo so we do not miss any files

kshyju:
> Just realized - this needs to updated in the other language composite files too - see https://github.com/Azure/azure-functions-docker/blob/dev/host/4/bullseye/amd64/java/java8/java8-composite.template#L4 - for example
> 
> Please do string search across the repo so we do not miss any files

Thank you! Fixed.

pragnagopa (approved):

PR #569, opened 2021-12-07
Update ReadMe - Add V4 tags. Remove V2 tags

pragnagopa:
Azure Functions V4 is now GA. Updating Readme to include new tags for V4. Removing V2 tags as Azure Functions V2 is deprecated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,3 +249,16 @@ Linux amd64 Tags
 | `3.0-slim`       | [Dockerfile](host/3.0/buster/amd64/dotnet/dotnet-slim.Dockerfile)       | Debian 10  |
 | `3.0-appservice` | [Dockerfile](host/3.0/buster/amd64/dotnet/dotnet-appservice.Dockerfile) | Debian 10  |
 
+# Contributing
+
+This project welcomes contributions and suggestions.  Most contributions require you to agree to a
+Contributor License Agreement (CLA) declaring that you have the right to, and actually do, grant us
+the rights to use your contribution. For details, visit https://cla.microsoft.com.
+
+When you submit a pull request, a CLA-bot will automatically determine whether you need to provide
+a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions
+provided by the bot. You will only need to do this once across all repos using our CLA.
+
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/).
+For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or
+contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
--- a/README.md
+++ b/README.md
@@ -12,22 +12,22 @@ CoreTools Healthcheck: [![CoreTools healthcheck](https://azure-functions.visuals
 
 Linux amd64 Tags
 
-| Tags                                       | Dockerfile                                                              | OS Version |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,6 +4,116 @@ CoreTools Healthcheck: [![CoreTools healthcheck](https://azure-functions.visuals
...
+
+#### Dotnet
+
+`mcr.microsoft.com/azure-functions/dotnet`
+
+Linux amd64 Tags
+
+| Tags                                       | Dockerfile                                                              | OS Version |
+|--------------------------------------------|-------------------------------------------------------------------------|------------|
+| `4`                                      | [Dockerfile](host/4/bullseye/amd64/dotnet/dotnet-inproc/dotnet.Dockerfile)            | Debian 10  |
```
pragnagopa: @balag0  / @kshyju  - please review - specifically tags, versions and OS version. Thanks!

pragnagopa:
cc @paulbatum @fabiocav as FYI

pragnagopa:
@ahmelsayed  / @balag0  - Docs https://hub.docker.com/_/microsoft-azure-functions-base need to be updated as well. Any idea where that content is pulled from?

README.md
```diff
@@ -4,6 +4,116 @@ CoreTools Healthcheck: [![CoreTools healthcheck](https://azure-functions.visuals
...
+| `4-slim`                                 | [Dockerfile](host/4/bullseye/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile)       | Debian 10  |
+| `4-appservice`, `4-dotnet3-appservice` | [Dockerfile](host/4/bullseye/amd64/dotnet/dotnet-inproc/dotnet-appservice.Dockerfile) | Debian 10  |
+| `4-dotnet3-core-tools` | [Dockerfile](host/4/bullseye/amd64/dotnet/dotnet-inproc/dotnet-core-tools.Dockerfile) | Debian 10  |
+
+
+`mcr.microsoft.com/azure-functions/dotnet-isolated`
+
+| Tags                                       | Dockerfile                                                              | OS Version |
+|--------------------------------------------|-------------------------------------------------------------------------|------------|
+| `4`                                      | [Dockerfile](host/4/bullseye/amd64/dotnet/dotnet-isolated/dotnet-isolated.Dockerfile)            | Debian 10  |
```
kshyju: I think this points to the old docker file (which we should delete from the repo to avoid confusion @CooperLink ). The actual docker file will be generated in the `out` directory when we run the `generate-composite.sh` script (example path: `\host\4\bullseye\amd64\out\dotnet\dotnet-isolated-appservice.Dockerfile`)

kshyju: This brings back the question of whether we should check-in the generated files (the `out` directory) to the source control. I feel like it might be a good idea to do so. If we want to check something quickly on the dockerfile we are using (ex: what version of extension bundle/aspnet sdk we are including in the final docker image produced).  Thoughts @CooperLink @pragnagopa

balag0: +1. In the last few weeks some customers were also looking for the final dockerfiles for reference (not just to use as a base image) to build their completely custom dockerfiles and having the final artifact directly available would definitely help.

pragnagopa: Final output docker files will be part of the release artifacts. Each release will have attached  assets - waiting for a release to ensure this flow works. For now, will remove the files from the table above ?

pragnagopa: DockerHub docs updated - https://hub.docker.com/_/microsoft-azure-functions-dotnet - could one of you approve this PR. I will have a follow up PR once we start generating docker files as part of next release.

kshyju (approved):

PR #579, opened 2021-12-13
[Python]Use bullseye instead for buster base images

### LINKED ISSUE #481: [V4 Images] V4 Python images are using python:3.9-slim-buster base images
There is no python:3.9-slim-bullseye image available currently.
Need to confirm if using the buster image is okay from functionality point of view.

+@vrdmr

pragnagopa:
Resolves #481

### DIFF
--- a/host/4/bullseye/amd64/python/python37/python37-composite.template
+++ b/host/4/bullseye/amd64/python/python37/python37-composite.template
@@ -30,7 +30,7 @@ RUN apt-get update && \
     apt-get install -y unixodbc msodbcsql17 mssql-tools && \
     # .NET Core dependencies
     apt-get install -y --no-install-recommends ca-certificates \
-    libc6 libgcc1 libgssapi-krb5-2 libicu63 libssl1.1 libstdc++6 zlib1g && \
+    libc6 libgcc1 libgssapi-krb5-2 libicu67 libssl1.1 libstdc++6 zlib1g && \
     rm -rf /var/lib/apt/lists/* && \
     # Custom dependencies:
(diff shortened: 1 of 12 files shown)


host/4/bullseye/amd64/python/python37/python37-composite.template
```diff
@@ -1,4 +1,4 @@
-FROM python:3.7-slim-buster
+FROM python:3.7-slim-bullseye
```
pragnagopa, line 1: @balag0 / @vrdmr  - any reason to use buster instead for bullseye?

balag0: Probably at that time due to - https://github.com/Azure/azure-functions-docker/issues/481

vrdmr: Yeah - thanks @Balag0 for pointing this out. This was due to buster not being available. This should be good to go now.

vrdmr (approved):
Thanks a lot @pragnagopa for the changes. LGTM

pragnagopa:
Taggging @TsuyoshiUshio as FYI

Tagging @CooperLink - as building App Services is failing.

pragnagopa:
@CooperLink - for PR builds - do we have artifacts that will show generated docker files?

CooperLink:
PR builds failing because bullseye is missing a dependency we had in buster it seems to me.  
'libicu63' is missing.  I will work on resolving this issue tomorrow. I think that I'll just need to add code to install that package.

PR #618, opened 2022-02-14
Enabling environmental variables from SSH

ffarinha-msft:
The proposed change aims to allow access to the environmental variables of Linux Function App from the SSH console, similar to what is possible on Azure Linux Web Apps.

The changes are based on startup script found on Linux Web Apps https://github.com/Azure-App-Service/ImageBuilder/blob/master/GenerateDockerFiles/dotnetcore/debian-9/init_container.sh


CooperLink:
This PR looks good to me. I don't know why the builds are failing but I am investigating

PR #619, opened 2022-02-14
Optimize DockerFiles for Layer Re-use

CooperLink:
Working to capture low hanging fruit to optimize docker layer re-use in platform.  Re-using docker layers makes downloading appservice images faster and cheaper. In this case, I was able to share a larger number of layers between images by moving copy directives.  Putting common Dockerfile portions near the top of our image directives encourages cache re-use for some of our largest layers.  For example, moving FunctionExtensionBundles to the first directive in our appservice images means that most images can share that data. 

The primary reason reducing unique image size is important is to support our image pre-caching step.  When the Linux roles start on platform, they cache several of our most used images.  During startup, they download those images so that when a customer's Function app needs an appservice image it is available without needing to download.  This means that caching images reduces cold start time.  By reducing the size of our docker images, we can cache more images in the space we have allotted and therefore reduce some cold starts. 

There should not be any functional changes to the images in this PR. Re-arranging the directives is all that is needed to benefit cache reuse. 

I built the images before and after implementing these changes to determine the impact.  I have pasted the results below.  One callout on the node 14 images as a representative.  
Unique Image Size
v3 1.212Gb -> 548Mb 55% decrease!
v4 591Mb -> 198.7mb 66% decrease!
image
image

image
image

### PR information

### DIFF
--- a/host/3.0/buster/amd64/dotnet/dotnet-isolated/dotnet-isolated-composite.template
+++ b/host/3.0/buster/amd64/dotnet/dotnet-isolated/dotnet-isolated-composite.template
@@ -5,9 +5,9 @@ FROM mcr.microsoft.com/dotnet/runtime:5.0
 ARG HOST_VERSION
 
 COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
-COPY sshd_config /etc/ssh/
 COPY start.sh /azure-functions-host/
 COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
+COPY sshd_config /etc/ssh/
 
 ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     HOME=/home \
--- a/host/3.0/buster/amd64/dotnet/dotnet-isolated/dotnet-isolated-composite.template
+++ b/host/3.0/buster/amd64/dotnet/dotnet-isolated/dotnet-isolated-composite.template
@@ -4,14 +4,17 @@ FROM mcr.microsoft.com/dotnet/aspnet:5.0 AS aspnet5
 FROM mcr.microsoft.com/dotnet/runtime:5.0
 ARG HOST_VERSION
 
+COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
+COPY sshd_config /etc/ssh/
+COPY start.sh /azure-functions-host/
+COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
+
 ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     HOME=/home \
     FUNCTIONS_WORKER_RUNTIME=dotnet-isolated \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
     HOST_VERSION=${HOST_VERSION}
 
-COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
-COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
 COPY --from=aspnet5 [ "/usr/share/dotnet", "/usr/share/dotnet" ]
 
 EXPOSE 2222 80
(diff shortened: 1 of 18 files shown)


host/3.0/buster/amd64/dotnet/dotnet-isolated/dotnet-isolated-composite.template
```diff
@@ -4,14 +4,17 @@ FROM mcr.microsoft.com/dotnet/aspnet:5.0 AS aspnet5
 FROM mcr.microsoft.com/dotnet/runtime:5.0
```
paulbatum, line 4: Not related to this PR, but it seems suss that this docker file references .NET 5, which goes end of life in 3 months. I will reach out to Fabio and Brett.

host/3.0/buster/amd64/node/node10/node10-composite.template
```diff
@@ -1,6 +1,10 @@
 FROM mcr.microsoft.com/dotnet/core/runtime-deps:3.1
 ARG HOST_VERSION
 
+COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
+COPY start.sh /azure-functions-host/
```
paulbatum, line 5: The `COPY sshd_config /etc/ssh/` was not moved up which breaks layer sharing. Is there a specific reason? (same is true for some other files in the PR but I won't repeat the question for those)

pragnagopa (approved):
💯 cc @paulbatum

pragnagopa:
Thanks for the PR @cooperlink ! Planning to send a separate PR for python?

host/4/bullseye/amd64/node/node14/node14-composite.template
```diff
@@ -26,9 +31,7 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
 RUN apt-get update && \
     apt-get install -y libc-dev
 
-COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
 COPY --from=runtime-image [ "/workers/node", "/azure-functions-host/workers/node" ]
```
paulbatum: Should we copy the node worker before we do setup_14.x / setup_16.x so that the worker layer is shared? Its probably not a huge saving, but it would also make it more consistent  with some of the other files that already have the worker at the top.

host/3.0/buster/amd64/node/node10/node10-composite.template
CooperLink: I was not able to figure out why.  But, for whatever reason the node builds fail when I move the ssh config information up.  When downloading openssh I get prompted for a user response regarding the config. Then the build stalls out and never succeeds. 
This only happens with node.  I will look into it today to see if I can resolve. IMO this portion doesnt matter as much as the layer for this sshd_config copy is only 352 B.  So the cache bust doesn't cost very much at all to us here.  However, it might make sense to move sshd down for all images so they can share the larger parts. ExtBund, Host etc.

host/4/bullseye/amd64/node/node14/node14-composite.template
CooperLink: Yeah I will do this now. thank you Paul

CooperLink: Updated node to share worker portion.  Saved another .3mb in shared sizes. Thanks Paul
image

host/3.0/buster/amd64/dotnet/dotnet-isolated/dotnet-isolated-composite.template
CooperLink: Resolving. Paul emailed.  Will work with him to update when necessary

host/3.0/buster/amd64/node/node10/node10-composite.template
CooperLink: Closing with no resolution here.  Going to track in WI that I may want to update the location of sshdconfig in all v3 images to promote best possible sharing.  But as mentioned above cache busting here is not a big deal.  The main thing that I want to do is make sure all v3 images can share the same host bits where able. For that reason I moved sshd below host in v3

CooperLink:
Merged.  Going to have discussion on Python, Powershell and Dotnet-OOP layer sharing. Larger work items not sure they will be complete by 3.5.0/4.2.0 release this week

pragnagopa:
cc @paulbatum

PR #625, opened 2022-02-17
Update powershell minor version tag

CooperLink:
Tested platform image selection code path to make sure that support for Powershell 7.2 was enabled. I noticed that platform is pulling from the location below to get the powershell7.2 image.  We are currently releasing images to the tag powershell:4-powershell**72**-appservice.  Blessed images must be published to the location specified by platform for support. 

Platform requests the following: 
`​mcr.microsoft.com/azure-functions/powershell:4-powershell7.2-appservice​` 

This PR updates the powershell versioning for 7.2 for platform support.  It also keeps versioning syntax the same with Python. Which is the other minor versioned language in this repository. 

### PR information

- [] New Unit tests were added for the changes made and CI is passing.

### DIFF
--- a/host/4/powershell-build.yml
+++ b/host/4/powershell-build.yml
@@ -50,7 +50,7 @@ steps:
                   host/4/bullseye/amd64/powershell/powershell70
       npm run test $IMAGE_NAME --prefix  test/
       docker push $IMAGE_NAME
-    displayName: powershell70
+    displayName: powershell7
     continueOnError: false
 
   - bash: |
@@ -61,7 +61,7 @@ steps:
                   host/4/bullseye/amd64/powershell/powershell70
       npm run test $IMAGE_NAME --prefix  test/
       docker push $IMAGE_NAME
-    displayName: powershell70-slim
+    displayName: powershell7-slim
     continueOnError: false
 
   - bash: |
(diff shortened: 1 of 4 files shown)


host/4/powershell-build.yml
```diff
@@ -50,7 +50,7 @@ steps:
                   host/4/bullseye/amd64/powershell/powershell70
       npm run test $IMAGE_NAME --prefix  test/
       docker push $IMAGE_NAME
-    displayName: powershell70
+    displayName: powershell7
```
Francisco-Gamino, line 53: Should this be `powershell70` as what we have in the path `host/4/bullseye/amd64/powershell/powershell70`? I am ok either way, but we should keep it consistent.

Francisco-Gamino (approved):
One minor comment. Otherwise, LGTM. Thanks!

host/4/powershell-build.yml
CooperLink: I chose to change this to reflect the image name instead of the folder structure because I think that the image name is more descriptive of what is being produced.  Will gladly change it back if you would prefer 70.  I considered making labels be  powershell 7.0 but, that doesn't make much sense since nothing else is named that way

Francisco-Gamino: Good point. Let's keep it as `PowerShell7`.

PR #627, opened 2022-02-24
Set Env Var ASPNET_CONTENTROOT to /azure-functions-host - to limit filewatchers

CooperLink:
Related to recent issue https://github.com/Azure/azure-functions-docker/issues/628

Update all images to add the ASPNET_CONTENTROOT variable to the environment.  This PR incorporates the changes introduced by hotfix releases 
https://github.com/Azure/azure-functions-docker/releases/tag/4.1.3.1
https://github.com/Azure/azure-functions-docker/releases/tag/3.4.2.1

A recent incident found that the Azure-Functions-Host spawned a very high number of inotify filewatchers on startup.  The cause was found to be a WebHostBuilder object in the Functions-Host code. The object [inherits ASPNET_CONTENTROOT as its context](https://github.com/aspnet/MetaPackages/blob/master/src/Microsoft.AspNetCore/WebHost.cs#L152).  Without that environment variable set, the object defaults to the `current directory` which is currently `/` for the linux images. 

This PR provides `ASPNET_CONTENTROOT = /azure-functions-host` to the WebHostBuilder which decreases the number of inotify filewatchers spawned on startup significantly.

### PR information

### DIFF
--- a/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
+++ b/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
@@ -10,7 +10,8 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     HOME=/home \
     FUNCTIONS_WORKER_RUNTIME=dotnet \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
-    HOST_VERSION=${HOST_VERSION}
+    HOST_VERSION=${HOST_VERSION} \
+    ASPNETCORE_ContentRoot=/azure-functions-host
 
 EXPOSE 2222 80
 
--- a/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
+++ b/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
@@ -11,7 +11,7 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     FUNCTIONS_WORKER_RUNTIME=dotnet \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
     HOST_VERSION=${HOST_VERSION} \
-    ASPNETCORE_ContentRoot=/azure-functions-host
+    ASPNETCORE_CONTENTROOT=/azure-functions-host
 
 EXPOSE 2222 80
 
--- a/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile
+++ b/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile
@@ -43,7 +43,7 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     FUNCTIONS_WORKER_RUNTIME=dotnet \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
     HOST_VERSION=${HOST_VERSION} \
-    ASPNETCORE_ContentRoot=/azure-functions-host
+    ASPNETCORE_CONTENTROOT=/azure-functions-host
 
 COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
 
--- a/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile
+++ b/host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile
@@ -42,7 +42,8 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     HOME=/home \
     FUNCTIONS_WORKER_RUNTIME=dotnet \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
-    HOST_VERSION=${HOST_VERSION}
+    HOST_VERSION=${HOST_VERSION} \
+    ASPNETCORE_ContentRoot=/azure-functions-host
 
 COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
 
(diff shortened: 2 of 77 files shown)


host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
```diff
@@ -10,7 +10,8 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     HOME=/home \
     FUNCTIONS_WORKER_RUNTIME=dotnet \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
-    HOST_VERSION=${HOST_VERSION}
+    HOST_VERSION=${HOST_VERSION} \
+    ASPNETCORE_ContentRoot=/azure-functions-host
```
pragnagopa: change this to all caps - https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host?view=aspnetcore-6.0#environment

pragnagopa:
@VpOfEngineering - please wait for this PR to be merged before starting release

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
```diff
@@ -10,7 +10,8 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
     HOME=/home \
     FUNCTIONS_WORKER_RUNTIME=dotnet \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
-    HOST_VERSION=${HOST_VERSION}
+    HOST_VERSION=${HOST_VERSION} \
```
balag0, line 13: I might have missed this in the emails, but why do we need HOST_VERSION?

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
```diff
@@ -10,7 +10,8 @@ ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
...
+    ASPNETCORE_CONTENTROOT=/azure-functions-host
```
balag0, line 14: From https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/web-host?view=aspnetcore-3.1#content-root, this should default to the folder where the app assembly resides.
Since we are not setting this currently , in these images shouldnt that already be /azure-functions-host ?

pragnagopa: very likely this is a bug - needs follow up with .net team - file watchers count drastically went down after setting this explicitly. @balag0  very like need to set this in mesh images as well until we figure this out in how to set this in functions host code.

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
pragnagopa: can you expand? this is an existing setting

balag0: sorry, i was using the wrong diff view :)

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
balag0: Just a quick guess (but would be good to validate), based on https://github.com/aspnet/MetaPackages/blob/master/src/Microsoft.AspNetCore/WebHost.cs#L152
if the env. variable is not set, it defaults to current directory.
In the dockerfiles there is no WORKDIR is set. so the current directory at that point could be /

If we have a way to repro/validate, can we add 
WORKDIR /azure-functions-host
to the dockerfiles and not set ASPNETCORE_CONTENTROOT explicitly
and see if that fixes the issue?

pragnagopa: good idea. @CooperLink - it is worth it to setup a repro for this on private stamp - host image without this change and Storage app setting - mount Azure Files with large folder structure.

CooperLink: After pushing hotfix changes to UD6 I will be setting up a private stamp repro for this to test the above. Thank you both for your comments

pragnagopa: @balag0  - @CooperLink  brought up a good point - would setting WorkingDir affect applications? I remember we considered this inthe past and decided was a breaking change - but OK to do this in V4 while it was in preview.

balag0: Yes, had a chat with Cooper just now as well. Yes agree that using the appsetting is the safer and better fix. Setting the WORKDIR was just an experiment to make sure we understand the behavior (in case there was a bug in the framework) :)

Also here is an approach to set this in the code (but dockerfiles is still better fix for I think)
https://stackoverflow.com/questions/59033413/c-sharp-aspnetcore-set-content-root-path

balag0 (approved):

pragnagopa:
@CooperLink - please update the PR description before merging.

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-composite.template
pragnagopa: @balag0 - is this change needed in mesh images?

balag0: Potentially since the setup is the same, but given we dont have any CRIs and the underlying issue here was exposed only after a config update on workers, I think its not urgent and we can wait until these images are rolled out and tested in case we identify any other scenarios to consider.

PR #726, opened 2022-06-23
Adding dotnet-isolated images using the .NET 7 Preview 5 bits

### LINKED ISSUE #725: Create images using dotnet 7 preview runtime bits
We need to create images using dotnet7 preview runtime so that customers can create isolated function apps using dotnet7 bits and deploy. These image will not replace the existing v4 images(using dotnet6 runtime bits), but will be additional images which will leave with the current v4 images.

kshyju:
Fixes #725

Adding dotnet-isolated images using the .NET 7 Preview 5 bits. Also updated the build & test pipeline.

Update:
Validation was done by the following steps:

 1. Created new function app via portal (dotnet runtime, dedicated linux hosting),
 2. Updated the app settings value for `FUNCTIONS_WORKER_RUNTIME` to `dotnet-isolated` via portal(Configuration blade)
 3. Added `LinuxFxVersion` setting with value `dotnet-isolated|7.0`
 4. Updated `LinuxFxVersion` config setting value via azure resource explorer to `dotnet-isolated|7.0`. Verified this updated value of `linux_fx_version` setting value via observer.
 5. Tried to access home page of the app.
 6. Checked the `WorkerLWASEventTable` table and verified that the infra was trying to pull the correct images

````
All("WorkerLWASEventTable")
| where PreciseTimeStamp >= datetime(2022-06-27T22:14:00.430193Z) and PreciseTimeStamp <= datetime(2022-06-27T22:15:00.430193Z)
| where Facility =~ "shkr-net7-isolated-linux-dedicated-cus"
| project PreciseTimeStamp, Facility, Msg
| where Msg startswith "Framework is:" or Msg startswith "Pulling image: mcr.microsoft.com/azure"
| distinct Msg
````
image

### DIFF
--- a/host/4/bullseye/amd64/dotnet/dotnet7-isolated/dotnet-isolated-composite.template
+++ b/host/4/bullseye/amd64/dotnet/dotnet7-isolated/dotnet-isolated-composite.template
@@ -1,0 +1,44 @@
+# Include ASP.NET Core shared framework from dotnet/aspnet image.
+FROM mcr.microsoft.com/dotnet/nightly/aspnet:7.0 AS aspnet7
+
+FROM mcr.microsoft.com/dotnet/nightly/runtime:7.0
+ARG HOST_VERSION
+
+COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
+COPY sshd_config /etc/ssh/
+COPY start.sh /azure-functions-host/
+COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
+
+ENV AzureWebJobsScriptRoot=/home/site/wwwroot \
+    HOME=/home \
+    FUNCTIONS_WORKER_RUNTIME=dotnet-isolated \
+    DOTNET_USE_POLLING_FILE_WATCHER=true \
+    HOST_VERSION=${HOST_VERSION} \
+    ASPNETCORE_CONTENTROOT=/azure-functions-host
+
+# Fix from https://github.com/GoogleCloudPlatform/google-cloud-dotnet-powerpack/issues/22#issuecomment-729895157
+RUN apt-get update && \
+    apt-get install -y libc-dev
+
+COPY --from=aspnet7 [ "/usr/share/dotnet", "/usr/share/dotnet" ]
+
+EXPOSE 2222 80
+
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends gnupg wget unzip curl dialog openssh-server && \
+    # Add remote dotnet debugger
+    curl -sSL https://aka.ms/getvsdbgsh | bash /dev/stdin -v vs2017u5 -l /root/vsdbg && \
+    echo "root:Docker!" | chpasswd
+
+# This is current not working with the .NET 6.0 image based on bullseye. Tracking here: https://github.com/Azure/azure-functions-docker/issues/451
+# Chrome headless dependencies
+# https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#chrome-headless-doesnt-launch-on-unix
+# RUN apt-get install -y xvfb gconf-service libasound2 libatk1.0-0 libc6 libcairo2 libcups2 \
+#    libdbus-1-3 libexpat1 libfontconfig1 libgbm1 libgcc1 libgconf-2-4 libgdk-pixbuf2.0-0 libglib2.0-0 \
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: test/index.ts [hunk 1]
```diff
@@ -23,6 +23,12 @@ const dotnetIsolated6 = {
   response: "Hello, Test"
 }
 
+const dotnetIsolated7 = {
+  package: `${storagePath}/dotnet-isolated7-functions.zip`,
```
kshyju, line 27: ~~I uploaded this zip with a test app. The test app is still using .NET 6 as TFM currently. Once [Update SDK for .NET 7 support](https://github.com/Azure/azure-functions-dotnet-worker/issues/935) is done, I will update the test app with the artifact produced from that work (the new SDK package to support .NET 7)~~

host/4/dotnet-build.yml
```diff
@@ -128,6 +128,48 @@ steps:
     displayName: dotnet-isolated-appservice (.NET 6)
     continueOnError: false
 
+  - bash: |
```
kshyju, line 131: FYI: I do not think we need [a core-tools version file](https://github.com/Azure/azure-functions-docker/blob/dev/host/4/bullseye/amd64/dotnet/dotnet-isolated/dotnet-isolated-core-tools.Dockerfile) with .NET7 changes, hence did not include that in this PR. Please correct me If I am wrong.

kshyju:
👋 @balag0 @CooperLink @pragnagopa Please review. Thank you!

cc @satvu FYI.

test/index.ts
kshyju: Updated the test app to use an SDK package which handles .NET 7. So the test app is a .NET 7 function app now.

host/4/dotnet-build.yml
```diff
@@ -128,6 +128,48 @@ steps:
...
+    continueOnError: false
+
+  - bash: |
+      set -e
+      IMAGE_NAME=azurefunctions.azurecr.io/azure-functions/4/dotnet-isolated:$(Build.SourceBranchName)-dotnet-isolated7.0-slim
+      IMAGE_ARRAY="$(image_list),$IMAGE_NAME"
+      echo "##vso[task.setvariable variable=image_list;]$IMAGE_ARRAY"
+
+      docker build -t $IMAGE_NAME \
+                  -f host/4/bullseye/amd64/dotnet/dotnet7-isolated/dotnet7-isolated-slim.Dockerfile \
```
CooperLink: Image name is dotnet-isolated-slim.  the 7 is causing build fail

host/4/bullseye/amd64/dotnet/dotnet7-isolated/dotnet-isolated-composite.template
```diff
@@ -0,0 +1,44 @@
+# Include ASP.NET Core shared framework from dotnet/aspnet image.
+FROM mcr.microsoft.com/dotnet/nightly/aspnet:7.0 AS aspnet7
+
+FROM mcr.microsoft.com/dotnet/nightly/runtime:7.0
```
CooperLink, line 4: Not a problem just a callout.   These images will not get layer sharing benefits other than the original host bits.  Sharing host bits will still  speed up build time. But, they will take up more unique space because they rely on a different base image. I understand this is necessary and its not an issue.

CooperLink (commented):
Please update pipeline for -slim image. Otherwise, LGTM thank you for your contribution Shyju

host/4/dotnet-build.yml
```diff
@@ -128,6 +128,48 @@ steps:
...
 
+  - bash: |
+      set -e
+      IMAGE_NAME=azurefunctions.azurecr.io/azure-functions/4/dotnet-isolated:$(Build.SourceBranchName)-dotnet-isolated7.0
+      IMAGE_ARRAY="$(image_list),$IMAGE_NAME"
+      echo "##vso[task.setvariable variable=image_list;]$IMAGE_ARRAY"
+
+      docker build -t $IMAGE_NAME \
+                  -f host/4/bullseye/amd64/dotnet/dotnet7-isolated/dotnet-isolated.Dockerfile \
+                  host/4/bullseye/amd64/dotnet/dotnet-isolated/
```
CooperLink: I think that the build context is wrong on these images. Non-appservice images should be built with their directory as context.  
To do this you will need to copy the shared config (start.sh and sshd_config) to this folder as well

host/4/dotnet-build.yml
kshyju: Thanks Cooper. Fixing this resolved the failing build steps.

CooperLink (approved):
LGTM thanks for applying changes.

cloudmelon:
@pragnagopa FYI :)

PR #731, opened 2022-06-29
Updating Python image ref to MCR GoldenImages

vrdmr:
Updating the Python image references from Docker.io to MCR golden images.

---

### PR information

### DIFF
--- a/host/3.0/buster/amd64/python/python37/python37.Dockerfile
+++ b/host/3.0/buster/amd64/python/python37/python37.Dockerfile
@@ -35,7 +35,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM python:3.7-slim-buster
+FROM mcr.microsoft.com/mirror/docker/library/python:3.7-buster
 ARG HOST_VERSION
 
 ENV LANG=C.UTF-8 \
--- a/host/3.0/buster/amd64/python/python37/python37.Dockerfile
+++ b/host/3.0/buster/amd64/python/python37/python37.Dockerfile
@@ -35,7 +35,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM mcr.microsoft.com/mirror/docker/library/python:3.7-buster
+FROM mcr.microsoft.com/mirror/docker/library/python:3.7-slim-buster
 ARG HOST_VERSION
 
 ENV LANG=C.UTF-8 \
--- a/host/3.0/buster/amd64/python/python38/python38.Dockerfile
+++ b/host/3.0/buster/amd64/python/python38/python38.Dockerfile
@@ -35,7 +35,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM mcr.microsoft.com/mirror/docker/library/python:3.8-buster
+FROM mcr.microsoft.com/mirror/docker/library/python:3.8-slim-buster
 ARG HOST_VERSION
 
 ENV LANG=C.UTF-8 \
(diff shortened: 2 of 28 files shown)


host/3.0/buster/amd64/python/python38/python38.Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM python:3.8-slim-buster
+FROM mcr.microsoft.com/mirror/docker/library/python:3.8-buster
```
CooperLink: Callout. Changed this image from slim to non-slim which will include more dependencies in the base image. Fine to keep as long as you are aware. 

Dependency set will only increase with this change so no risk of breaking apps.

host/3.0/buster/amd64/python/python37/python37.Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM python:3.7-slim-buster
+FROM mcr.microsoft.com/mirror/docker/library/python:3.7-buster
```
CooperLink: Same thing here. Likely the same for all images. Won't callout others. 
Previously this relied on Slim. Now it relies on fat

host/3.0/buster/amd64/python/python39/python39-composite.template
```diff
@@ -1,4 +1,4 @@
-FROM python:3.9-slim-buster
+FROM mcr.microsoft.com/mirror/docker/library/python:3.9-slim-buster
```
CooperLink, line 1: I don't like that appservice images rely on slim images but, non-appservice images (no suffix like Python39.Dockerfile) rely on fat images.

Appservice and nonsuffix images should be identical other than SSH tooling that we add to these images. 
I think that you should keep everything slim for the time being. I understand that its non-intuitive. sorry

host/3.0/buster/amd64/python/python37/python37.Dockerfile
CooperLink: Changed my mind. Please change this to -slim to minimize changes. Will keep parity between this and appservice images. Please do this for all non-suffix images. Ideally in a single commit

host/3.0/buster/amd64/python/python39/python39.Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN EXTENSION_BUNDLE_VERSION=1.8.1 && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM python:3.9-slim-buster
+FROM mcr.microsoft.com/mirror/docker/library/python:3.9-buster
```
CooperLink: Make -slim. Won't comment on other images. But please apply this comment to all nonsuffix images.

host/4/bullseye/amd64/python/python310/python310-buildenv.Dockerfile
```diff
@@ -3,7 +3,7 @@
 # Licensed under the MIT License. See https://go.microsoft.com/fwlink/?linkid=2090316 for license information.
 #-------------------------------------------------------------------------------------------------------------
 
-FROM python:3.10-slim-bullseye
+FROM mcr.microsoft.com/mirror/docker/library/python:3.10-slim
```
CooperLink, line 6: Bullseye Suffix image does not exist. 
This is not ideal but is not our responsibility. Just calling out.

host/4/bullseye/amd64/python/python310/python310.Dockerfile
```diff
@@ -34,7 +34,7 @@ RUN apt-get update && \
     rm -f /$EXTENSION_BUNDLE_FILENAME_V3 &&\
     find /FuncExtensionBundles/ -type f -exec chmod 644 {} \;
 
-FROM python:3.10-slim-bullseye
+FROM mcr.microsoft.com/mirror/docker/library/python:3.10-slim
```
CooperLink, line 37: This image stayed slim.  Which is good. Please keep this consistency in all non-suffix images.

CooperLink (changes requested):
Requesting changes to v3 images.  Please keep non-suffix images on the slim base image. Reduces the changeset here and  provides parity between non-suffix and -appservice images

CooperLink (approved):

PR #802, opened 2022-10-31
Nightly Build and Publish pipelines for v4 images

CooperLink:
This PR should introduce a new nightly build pipeline that triggers a nightly release flow. The nightly build pipeline runs every night at midnight UTC and merges the newest bits on dev into a branch -> nightly-build then pushes that to remote. Nightly-build has been added as a trigger for all v4 build pipelines. So, the push to nightly-build should trigger builds of all images. Finally, the publish pipeline has been updated with a 3 AM cron job that should run after all builds have been finished.  When the pipeline is triggered using the cron job, it targets the 'nightly-build' images and pushes them to publish on a 4-nightly tag schema

This PR also updates build pipelines in v4 with better re-fresh pipeline support. 

Releasing new Host versions.
[V3 Release](PASTE_V3_LINK_HERE)
[V2 Release](PASTE_V2_LINK_HERE)

---

### PR information

- [x] The title of the PR is clear and informative.
- [x] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/development-docs/cleaning-up-commits.md).
- [ ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ ] New Unit tests were added for the changes made and CI is passing.

PR #804, opened 2022-11-04
Revert "Nightly Build and Publish pipelines for v4 images"

CooperLink:
Reverts Azure/azure-functions-docker#802

Need to find a different way to trigger the build pipelines. Azure Devops only supports a single trigger definition


VpOfEngineering (approved):
mtgl

PR #810, opened 2022-11-21
Simplify rebuild pipeline to gh action

CooperLink:
Azure Pipelines Rebuild.Yml is not necessary to trigger the nightly-build branch update. This GH action should update the Nightly-Build branch from Dev every day.  

Each day at 12 UTC, GH action should take the current dev branch. Update it with a DateTime string under build-time.txt and force push to the nightly-build branch.  This push will trigger all of the build pipelines trigger rebuilds of all of our images. Then, the publish pipeline should be run each night at about 3 AM UTC and push the nightly-build images to the :4-nightly tags. 

Still working out CI/CD kinks with this so I will closely monitor pipelines to get nightly's up and running

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -20,6 +20,9 @@ jobs:
         set -e
         git config --local user.email "azfuncgh@github.com"
         git config --local user.name "Azure Functions"
+        date > build-time.txt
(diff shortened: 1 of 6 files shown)


VpOfEngineering (approved):
LG,TM

PR #823, opened 2022-12-21
Update Refresh Images Github Action

CooperLink:
The refresh pipelines are having a hard time automatically triggering builds for v3/v4 images. The wildcard that should be matching refresh branches is not successfully operating.  This seems to be a bug with Azure Devops. I was able to override the triggers from Azure Devops UI and match the refresh branch with wildcards as expected.

This PR moves the wildcard trigger to away from the bottom of the trigger list and partially simplifies v3 pipelines. I am not sure that this will resolve the wildcard matching bug, but it is a safe change and I will continue to improve these pipelines.


TsuyoshiUshio (approved):
Approve for testing CI Pipeline.

PR #835, opened 2023-01-18
Update readme. Add quickstart tags. Remove deprecated V3 tags.

pragnagopa:
- Update Read Me to include dotnet-isolated 7.0 tags , remove V3 tags , Add quickstart tags
- Remove Powershell 7 from build and publish pipeline. Powershell 7.0 is not supported on V4
- MCR Docs update PR - https://github.com/microsoft/mcrdocs/pull/32024

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Linux amd64 Tags
 | `4-slim`                                 | Debian 11  |
 | `4-appservice` | Debian 11  |
 | `4-dotnet6-core-tools` |  Debian 11  |
+| `4-appservice-quickstart` | Debian 11 |
 
 
 `mcr.microsoft.com/azure-functions/dotnet-isolated`
@@ -29,7 +30,13 @@ Linux amd64 Tags
 | `4`                                      | Debian 11  |
 | `4-dotnet-isolated5.0-slim`                               | Debian 11 |
 | `4-appservice`, `4-dotnet-isolated5.0-appservice` |Debian 11  |
+| `4-dotnet-isolated6.0` | Debian 11  |
 | `4-dotnet-isolated6.0-core-tools` | Debian 11  |
+| `4-dotnet-isolated6.0-appservice-quickstart` | Debian 11 |
+| `4-dotnet-isolated7.0` | Debian 11  |
+| `4-dotnet-isolated7.0-core-tools` | Debian 11  |
+| `4-dotnet-isolated7.0-appservice-quickstart` | Debian 11 |
+
 
 #### Node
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -20,6 +20,7 @@ Linux amd64 Tags
 | `4-slim`                                 | Debian 11  |
 | `4-appservice` | Debian 11  |
 | `4-dotnet6-core-tools` |  Debian 11  |
+| `4-appservice-quickstart` | Debian 11 |
```
CooperLink, line 23: In my opinion we should consider not listing the quickstart images as they are equivalent to the -appservice images.  We don't want customers to arbitrarily pin to the quickstart images thinking they are getting some benefit from it. Rather have all customers targetting the same appservice images. Makes hotfixes, future changes easier.

host/4/publish.yml
```diff
@@ -295,36 +295,6 @@ steps:
...
-
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)-slim
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)-appservice
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)-appservice-quickstart
-
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)-powershell7
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)-powershell7-slim
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)-powershell7-appservice
-      docker push $TARGET_REGISTRY/powershell:$(TargetVersion)-powershell7-appservice-quickstart
```
CooperLink, line 322: Would you like me to delete the existing powershell 7 images? Or keep them up to support customers who are still on them ?

CooperLink (approved):
2 comments.  Mostly looks good to me. I think we should not list the quickstart images as they are just renamed appservice images. I would prefer no more customers depending on those. So we can explore getting rid of that clutter in the future

host/4/publish.yml
pragnagopa: unfortunately we cannot remove as the images are public but we can stop building them as it is not a supported version

README.md
pragnagopa: quickstart images will not be listed on docker hub - https://hub.docker.com/_/microsoft-azure-functions but it is worth calling out all the images that are being built in this repo in the read-me for better visibility.

pragnagopa: Resolving the comment and merging the PR. @Cooperlink - lets sync up offline if you have more concerns

PR #1000, opened 2023-11-15
Create Host Version and Extension Bundles Validation Script

CooperLink:
Recently we had Mariner images fall off of the Host Version wagon and get pinned accidentally to 4.26.0.  I wrote this script to force a future release to fail if any Dockerfile's are referencing a different HostVersion than the first Dockerfile picked up by the script.  That means that per Major version. Every image has have the same Host Version to allow us to release anything. 

EDIT : Added another script to test Extension Bundles version 2,3 4 in both host release versions. This one will fail v3 builds also!

NOTE : THIS WILL CAUSE V3 DEPLOYMENT TO FAIL AS WE HAVE OUT OF BAND VERSIONS IN THAT RELEASE.  This can be manually overridden in a situation where it is necessary. 

image
image
image

### DIFF
--- a/.github/workflows/autopr.yaml
+++ b/.github/workflows/autopr.yaml
@@ -26,6 +26,7 @@ jobs:
         ./verify-pipelines.sh ${{ github.event.inputs.majorVersion }}
         ./update-version.sh ${{ github.event.inputs.majorVersion }} ${{ github.event.inputs.targetVersion }}
         ./validate-hostversion.sh ${{ github.event.inputs.majorVersion }}
+        ./validate-extbundle.sh ${{ github.event.inputs.majorVersion }}
         git add .
         git commit -m "Update Version ${{ github.event.inputs.targetVersion }}"
     - name: Create PullRequest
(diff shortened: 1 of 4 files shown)


.github/workflows/autopr.yaml
```diff
@@ -25,6 +25,8 @@ jobs:
         cd host
         ./verify-pipelines.sh ${{ github.event.inputs.majorVersion }}
         ./update-version.sh ${{ github.event.inputs.majorVersion }} ${{ github.event.inputs.targetVersion }}
+        ./validate-hostversion.sh ${{ github.event.inputs.majorVersion }}
```
pragnagopa, line 30: For hotfixes that are specific to a single image - this check needs to be skipped - can you add comment in this file and also, update https://github.com/Azure/azure-functions-docker/wiki/Release-Images-for-Linux-Dedicated-(GitHub-Action) to include how to skip this check?

pragnagopa (dismissed):
Update docs before merging to ensure this check does not block hotfix releases.

.github/workflows/autopr.yaml
CooperLink: Added thank you

CooperLink:
Docs were updated to indicate that these scripts will cause release to fail if host /ext bundle versions do not match. And that they can be safely removed if the intention is to update only a small subset of imaegs.

amamounelsayed (approved):

PR #1005, opened 2023-11-16
Add deprecation warning for v3 images

zahalzel:
Work item: https://msazure.visualstudio.com/Antares/_workitems/edit/25774564/

Releasing new Host versions.

---

Adds the following deprecation message to the startup script for v3 images:

```
**WARNING**: You are using an outdated version of the Azure Functions runtime:
https://learn.microsoft.com/en-us/azure/azure-functions/migrate-version-3-version-4
====================================================================================
```

image

### PR information

- [ x] The title of the PR is clear and informative.
- [ x] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/development-docs/cleaning-up-commits.md).
- [x ] If applicable, the PR references the bug/issue that it fixes in the description.
- [ x] New Unit tests were added for the changes made and CI is passing.


host/3.0/buster/amd64/dotnet/dotnet-isolated/start.sh
```diff
@@ -21,6 +21,11 @@ sed -i "s/SSH_PORT/$SSH_PORT/g" /etc/ssh/sshd_config
 
 service ssh start
 
+echo -e "**WARNING**: You are using an outdated version of the Azure Functions runtime:
+https://learn.microsoft.com/en-us/azure/azure-functions/migrate-version-3-version-4
```
pragnagopa: any reason for not reusing /azure-functions-host/start_deprecated.sh here?

zahalzel: Made some changes to see if we can call the start_deprecated.sh script from the start.sh script in order to consolidate the warning message :)

pragnagopa (dismissed):
Wait for approval from @Cooperlink as well.

CooperLink (approved):
lgtm

arroyc:
why do we need the start_deprecate.sh in each of the language folder? it looks like same script to me? May be we can keep only single copy of it?

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile
```diff
@@ -56,4 +56,8 @@ COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
 
 COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
 
-CMD [ "/azure-functions-host/Microsoft.Azure.WebJobs.Script.WebHost" ]
+COPY start_deprecated.sh /azure-functions-host/
+
+RUN chmod +x /azure-functions-host/start_deprecated.sh
+
+CMD ["/azure-functions-host/start_deprecated.sh"]
```
arroyc, line 63: are we overriding this command later stages? what is the reason of using CMD here instead of RUN?

host/3.0/buster/amd64/dotnet/dotnet-inproc/start_deprecated.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+echo -e "**WARNING**: You are using an outdated version of the Azure Functions runtime:
+https://learn.microsoft.com/en-us/azure/azure-functions/migrate-version-3-version-4
+===================================================================================="
```
arroyc, line 4: may be we can keep only this much in a new script, also i think once we are done with running the script in the build phase, we should remove it as well

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile
CooperLink: CMD and RUN work differently. https://stackoverflow.com/questions/37461868/difference-between-run-and-cmd-in-a-dockerfile

host/3.0/buster/amd64/dotnet/dotnet-inproc/start_deprecated.sh
CooperLink: CMD is basically the entrypoint of the script. It runs when the container is started and therefore we have to start the host at the end of any startup script

CooperLink:
> why do we need the start_deprecate.sh in each of the language folder? it looks like same script to me? May be we can keep only single copy of it?

Single copy of this file would require us to refactor the build pipelines to handle the changed context. Out of scope for v3 images totally.

host/3.0/buster/amd64/dotnet/dotnet-inproc/dotnet-slim.Dockerfile
arroyc: my bad .. i was thinking this is only needed during build phase .. ignore this comment

host/3.0/buster/amd64/dotnet/dotnet-inproc/start_deprecated.sh
arroyc: yes i understood this is sort of entrypoint for the container, ignore this

PR #1060, opened 2024-04-05
Update dotnet8  EXPOSE port

CooperLink:
Common Docker image use cases take dependency on the EXPOSE docker configuration.  As part of the transition to dotnet8, we lost automatic inheritance of EXPOSE 80 and picked up EXPOSE 8080 from the new Dotnet 8 images. This causes unwanted issues with custom images for customers using dotnet8. The appservice variant of this image already exposed 80. This just takes that change to all dn8 images


VpOfEngineering (approved):
LGTM
GTML
TMLG
MLGT

PR #1087, opened 2024-05-15
Add appservice tag for dedicated .NET 8 in proc image

aishwaryabh:
This PR adds an appservice tag to the `publish.yml` and `republish.yml` files so that the tooling feed can consume these changes.

### PR information

- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/development-docs/cleaning-up-commits.md).
- [x] New Unit tests were added for the changes made and CI is passing.

### DIFF
--- a/host/4/publish.yml
+++ b/host/4/publish.yml
@@ -80,8 +80,13 @@ steps:
   - bash: |
       set -e
       docker pull $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-dotnet8
+      docker pull $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-dotnet8-appservice
+
       docker tag $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-dotnet8 $TARGET_REGISTRY/dotnet:$(TargetVersion)-dotnet8
+      docker tag $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-dotnet8-appservice $TARGET_REGISTRY/dotnet:$(TargetVersion)-dotnet8-appservice
+
       docker push $TARGET_REGISTRY/dotnet:$(TargetVersion)-dotnet8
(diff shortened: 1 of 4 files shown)


host/4/publish.yml
```diff
@@ -80,8 +80,13 @@ steps:
   - bash: |
       set -e
       docker pull $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-dotnet8
+      docker pull $SOURCE_REGISTRY/dotnet:$(PrivateVersion)-dotnet8-appservice
```
pragnagopa: @Cooperlink should this be 8.0?

CooperLink: I do not know it depends on what the platform is pulling for this stack. That is a good callout. I can check the code/ look for logs/ find an old linux dedicated stamp that is kicking around maybe

CooperLink: This should be 8.0 to match the expected tag in Antares. Thank you Pragna

mattchenderson: Per discussion on Teams, it seems there would not be a difference in behavior from that, and other work will be needed to account for whatever form the tag takes. Therefore, the current form here should be equally fine.

aishwaryabh: I went ahead and updated all the tags to end with 8.0 (including the existing images)

pragnagopa (changes requested):
Lets wait for ANT changes to support .Net 8 inproc

pragnagopa:
addressed in https://github.com/Azure/azure-functions-docker/pull/1096 ; closing this PR

@Fabiocav @CooperLink  @aishwaryabh  @mattchenderson

PR #1116, opened 2024-07-19
Update the pipeline, dotnet8 builds

CooperLink:
Building dotnet8 is broken because of an issue with merging in May.  This updates the state of the dotnet8 images to include the shell scripts that it needs. It also sets up the node republish pipeline to work better.


zahalzel (dismissed):

TsuyoshiUshio (approved):
LGTM

PR #1144, opened 2024-08-19
Fix dotnet-inproc-appservice build

pragnagopa:


host/4/bullseye/python/python37/python37-composite.template
```diff
@@ -62,8 +62,6 @@ COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
 
 ENV FUNCTIONS_WORKER_RUNTIME_VERSION=3.7
 
-EXPOSE 2222 80
```
CooperLink, line 65: if we remove these expose rules then SSH won't work on Python3.7

host/4/bullseye/python/python38/python38-composite.template
```diff
@@ -61,8 +61,6 @@ COPY --from=runtime-image [ "/FuncExtensionBundles", "/FuncExtensionBundles" ]
 
 ENV FUNCTIONS_WORKER_RUNTIME_VERSION=3.8
 
-EXPOSE 2222 80
```
CooperLink, line 64: Removing expose removes ssh

host/4/mariner/dotnet-isolated/dotnet7-isolated-mariner.Dockerfile
```diff
@@ -37,5 +38,6 @@ RUN dnf install -y glibc-devel
 
 COPY --from=runtime-image [ "/azure-functions-host", "/azure-functions-host" ]
 COPY --from=aspnet7 [ "/usr/share/dotnet", "/usr/share/dotnet" ]
+COPY --from=dn8-sdk-image [ "/usr/share/dotnet", "/usr/share/dotnet" ]
```
CooperLink, line 41: Why do we need to include the dotnet8 sdk image

pragnagopa: without this running the docker image is failing. I merged the PR as the changes are close to green build. @CooperLink - I think its good for you take a final pass after I get all the green builds to see if there is any other fixes we need to do!

PR #1172, opened 2024-10-01
Update nonappservice files to selfcontain the host

CooperLink:
Build host to be selfcontained

[Breaking change: Runtime-specific apps no longer self-contained - .NET | Microsoft Learn](https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/8.0/runtimespecific-app-default)


CooperLink (commented):
Observed all changes. No obvious issues. Mariner is probably pretty big/ installs dn8 sdk but I am not going to adjust it here.

pragnagopa (approved):

host/4/dotnet-isolated-build.yml
```diff
@@ -114,7 +114,7 @@ jobs:
         displayName: $(dotnetVersion)
         continueOnError: false
   - job: Publish
-    dependsOn: Dotnet
+    dependsOn: Dotnet_iso
```
pragnagopa: this throws --change to dotnet

pragnagopa (changes requested):
pipline changes

host/4/dotnet-isolated-build.yml
CooperLink: Are you sure? In the previous PR i included the other change but not this change on line 117. I can remove this change but i think part of the reason that this does not run on PRs is because the name

CooperLink: Removed them anyway

pragnagopa (approved):

PR #1175, opened 2024-10-11
Added opentelemetry support for python 3.12

gavin-aguiar:
Adding opentelemetry distro for python 3.12. This is a part of supporting opentelemetry by default for python. 
Starting this with python 3.12 since this will be in public preview soon.  Will add it to the rest of the python version after we monitor and get feedback from the community.

Releasing new Host versions.

---

### PR information

- [ ] The title of the PR is clear and informative.
- [ ] There are a small number of commits, each of which has an informative message. This means that previously merged commits do not appear in the history of the PR. For information on cleaning up the commits in your pull request, [see this page](https://github.com/Azure/azure-powershell/blob/dev/documentation/development-docs/cleaning-up-commits.md).

### DIFF
--- a/host/4/bookworm/python/python312/python312-appservice.Dockerfile
+++ b/host/4/bookworm/python/python312/python312-appservice.Dockerfile
@@ -75,6 +75,11 @@ COPY --from=python [ "/opt", "/opt" ]
 RUN for file in /opt/python/3.12/bin/*; do \
         ln -sf "$file" /usr/bin/$(basename "$file"); \
     done
+RUN ln -s /opt/python/3.12/lib/libpython3.12.so.1.0 /usr/lib/libpython3.12.so.1.0
+
+# Install opentelemetry packages
+RUN pip install azure-monitor-opentelemetry-exporter
+RUN pip install azure-monitor-opentelemetry
 
 ENV LANG=C.UTF-8 \
     ACCEPT_EULA=Y \ 
@@ -86,7 +91,6 @@ ENV LANG=C.UTF-8 \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
     HOST_VERSION=${HOST_VERSION} \
     ASPNETCORE_CONTENTROOT=/azure-functions-host \
-    LD_LIBRARY_PATH=/opt/python/3.12/lib:$LD_LIBRARY_PATH \
     FUNCTIONS_WORKER_RUNTIME_VERSION=3.12 
 
 RUN apt-get update && \
--- a/host/4/bookworm/python/python312/python312-appservice.Dockerfile
+++ b/host/4/bookworm/python/python312/python312-appservice.Dockerfile
@@ -78,8 +78,7 @@ RUN for file in /opt/python/3.12/bin/*; do \
 RUN ln -s /opt/python/3.12/lib/libpython3.12.so.1.0 /usr/lib/libpython3.12.so.1.0
 
 # Install opentelemetry packages
-RUN pip install azure-monitor-opentelemetry-exporter
-RUN pip install azure-monitor-opentelemetry
+RUN pip install azure-monitor-opentelemetry-exporter azure-monitor-opentelemetry
 
 ENV LANG=C.UTF-8 \
     ACCEPT_EULA=Y \ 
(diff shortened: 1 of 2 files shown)


host/4/bookworm/python/python312/python312-appservice.Dockerfile
```diff
@@ -75,6 +75,11 @@ COPY --from=python [ "/opt", "/opt" ]
 RUN for file in /opt/python/3.12/bin/*; do \
         ln -sf "$file" /usr/bin/$(basename "$file"); \
     done
+RUN ln -s /opt/python/3.12/lib/libpython3.12.so.1.0 /usr/lib/libpython3.12.so.1.0
+
+# Install opentelemetry packages
+RUN pip install azure-monitor-opentelemetry-exporter
```
CooperLink: Can you do this in a single statement

host/4/bookworm/python/python312/python312-appservice.Dockerfile
```diff
@@ -86,7 +91,6 @@ ENV LANG=C.UTF-8 \
     DOTNET_USE_POLLING_FILE_WATCHER=true \
     HOST_VERSION=${HOST_VERSION} \
     ASPNETCORE_CONTENTROOT=/azure-functions-host \
-    LD_LIBRARY_PATH=/opt/python/3.12/lib:$LD_LIBRARY_PATH \
```
CooperLink, line 89: Is this not necessary? I can't remember why it was added

host/4/bookworm/python/python312/python312-appservice.Dockerfile
gavin-aguiar: Done

host/4/bookworm/python/python312/python312-appservice.Dockerfile
gavin-aguiar: We don't need it. Tested it without this variable.
We do need the libpython3.12.so.1.0 which I have linked.

CooperLink: Could you explain the thought process behind this change?  If it works it works but IMO this is worse than using the builtin LD LIB PATH.  Using a symlink to the SO is probably fine but if there is anything else in that path that is used in the dir we might need it could be hard to tell until we hit a runtime error that its necessary

image


For instance in this image we see libpython3.so  its tiny and probably used nowhere. But if its part of some vague path thats only called in some situations we would encounter this issue only then. Just raising this as a concern. If you are confident I am fine with keeping it

host/4/bookworm/python/python312/python312.Dockerfile
```diff
@@ -73,6 +73,10 @@ COPY --from=python [ "/opt", "/opt" ]
 RUN for file in /opt/python/3.12/bin/*; do \
         ln -sf "$file" /usr/bin/$(basename "$file"); \
     done
+RUN ln -s /opt/python/3.12/lib/libpython3.12.so.1.0 /usr/lib/libpython3.12.so.1.0
+
+# Install opentelemetry packages
+RUN pip install azure-monitor-opentelemetry-exporter azure-monitor-opentelemetry
```
CooperLink, line 79: Do you need to add any flags like -y for pip? Just wondering if this could cause build to hang waiting for input at some time.

host/4/bookworm/python/python312/python312-appservice.Dockerfile
gavin-aguiar: There are 3 reasons that I made this change:
1) While building this image with the open telemetry package, I got an error about a missing libpython3.12.so dependency. That suggest we need this to install packages to the image.
2) The python 3.12 consumption image has the same symlink. Thats what I got it from. Better to have it in both images for consistency.
3) The LD LIB PATH was not getting used. After building the image, there was a warning message for this line, saying its not used. Since it doesn't have any affect, might as well remove it from the image.

host/4/bookworm/python/python312/python312.Dockerfile
gavin-aguiar: pip install does not have a non interactive mode like apt get. I dont think it would hang at this point. Either it will install or throw an error.

host/4/bookworm/python/python312/python312-appservice.Dockerfile
CooperLink: We should only drive consistency across the images with the right approach. With that being said I think this is fine approving

CooperLink (approved):

host/4/bookworm/python/python312/python312-appservice.Dockerfile
gavin-aguiar: Thanks. Just wanted to point out that this is the right approach. Adding the LD LIB PATH was a no-op. Linking the  libpython was needed since now all the python libraries are in opt which is not the pattern we follow in functions. 
This change will be consistent with what we have in our consumption images,

